PR #5, opened 2019-09-09
Device contexts and events.

mdavis36:
Type erased device implementation based off of the implementation by @trws . This implementation includes type-erased device contexts and type erased event objects. The device context now implements a unique pointer to the underlying object and a get<>() function to dynamically cast back to the original type.


mdavis36:
@trws do you think it would be better to split device contexts and events into separate .hpp files?

include/camp/device.hpp
```diff
@@ -0,0 +1,313 @@
...
+    cuda = 8,
+    // omp_target = 16, not sure this is a meaningful difference
+    hip = 32
+  };
+
+  class CudaEvent
+  {
+    public:
+      CudaEvent(){ cudaEventCreateWithFlags(&m_event, cudaEventDisableTiming); }
+      void capture(cudaStream_t stream){ cudaEventRecord(m_event, stream); }
```
davidbeckingsale: I thought we were going to drop this function?

mdavis36: You're right I'll take this out :)

davidbeckingsale:
I think this is looking good. The only changes we might want to make are are in the naming/style to keep it consistent with the rest of camp. I'll let @trws weigh in on that.

include/camp/device.hpp
```diff
@@ -0,0 +1,311 @@
...
+  };
+
+  class Cuda 
+  {
+    static cudaStream_t get_a_stream(int num)
+    {
+      // TODO consider pool size
+      static cudaStream_t streams[16] = {};
+      static int previous = 0;
+      // TODO deal with parallel init
```
trws: I realize I'm the one that left this in here, but it would be good to squash this TODO out before the merge.  A moving the for-loop into a lambda invoked by [std::call_once](https://en.cppreference.com/w/cpp/thread/call_once) ought to take care of it.

include/camp/device.hpp
```diff
@@ -0,0 +1,311 @@
...
+      void memcpy(void *dst, const void *src, size_t size)
+      {
+	m_value->memcpy(dst, src, size);
+      }
+      void memset(void *p, int val, size_t size) { m_value->memset(p, val, size); }
+      Event get_event() { return m_value->get_event(); }
+      void wait_on(Event *e) { m_value->wait_on(e); }
+
+    private:
+      class ContextConcept {
```
trws: I'd prefer we call this ContextBase or ContextInterface or something like that.  Concept implies it as a checked concept for a class of types, and I actually want to add a context concept we can use for such checks to this code at some point.

include/camp/device.hpp
```diff
@@ -0,0 +1,311 @@
...
+	public:
+	  EventModel(T const& modelVal) : m_modelVal(modelVal) {}
+	  bool check() const override { return m_modelVal.check(); }
+	  void wait() const override { m_modelVal.wait(); }
+	  T *get() { return &m_modelVal; }
+	private:
+	  T m_modelVal;
+      };
+
+      std::unique_ptr<EventConcept> m_value;
```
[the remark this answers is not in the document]
mdavis36: I must have misunderstood, and thought we wanted it to not be copy constructable? However I guess it makes sense if it accidentally gets destroyed, and chai/umpire is still trying to handle  memory related to the object. I'll convert it to a value_ptr.

davidbeckingsale:
@trws - can you take another look at this?

include/camp/device.hpp
```diff
@@ -0,0 +1,273 @@
...
+    }
+    void free(void *p) { free(p); }
+    void memcpy(void *dst, const void *src, size_t size)
+    {
+      memcpy(dst, src, size);
+    }
+    void memset(void *p, int val, size_t size) { std::memset(p, val, size); }
+  };
+
+  class Context
```
mdavis36: We've changed the name of the underlying base class from "Device" to "Context", do we want to rename the namespace form "devices" to something else?  @trws

trws: Sure, how about "resource" or "resources"?  @davidbeckingsale, thoughts?

trws:
I'd like to merge a version of this this week, so we can get moving on it here and elsewhere, it looks like it's in pretty good shape for a v1.  If possible, lets do the rename, then I'd like the whole thing wrapped in an inline namespace so we can do a slow fail-over if we decide we need to (perhaps being overly cautious, but I'd rather have it than not and find we want it).


Something like this:

```c++
namespace resource {
inline namespace v1 {
// stuff
}
}
```

include/camp/device.hpp
davidbeckingsale: Do we need to determine between an execution resource and another kind of resource?

trws: Perhaps?  Each of these has both compute and memory associations, so it's kinda both.

mdavis36:
@trws I think we're ready to merge what is here now, could you look over some of the latest changes please. I needed to add a config header for integration with CHAI and Umpire and want to make sure you're okay with this.

CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ endif()
 add_library (camp INTERFACE)
 target_include_directories (camp INTERFACE
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+  $<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>
```
davidbeckingsale: Is this still needed?

mdavis36:
@trws Referring to your previous comment on inline namespaces, those have been added.

trws:
This one has gone long enough, and it looks good to me, so I'm going to merge.  That said @mdavis36 would you put up a new PR to change contexts.hpp to resources.hpp to make it match the namespaces?

Good work man, thanks for following it through!

PR #10, opened 2019-11-14
Feature/hip

noelchalmers:
Adding support for AMD's HIP in camp. 

There is currently a compiler bug which causes link time errors when using camp::tuples. There is an imperfect work around in place, but it will likely cause issues using host-side reductions in RAJA. We're tracking this issue and will update when a fix has been implemented.


trws:
Thanks @noelchalmers.

PR #11, opened 2019-11-15
Add make_unique

davidbeckingsale:
As discussed, adding a make_unique to camp for use in other projects


trws:
Excellent, thanks @davidbeckingsale!  Only request, add an include of this into the top-level camp.hpp.

davidbeckingsale:
No problem! Having it here is a good idea.

PR #12, opened 2019-12-10
Renaming context.hpp to resource.hpp

mdavis36:
Renaming context.hpp to resource.hpp to match namespace name.

Adds the copyright information to all of the resource files.


trws:
Works for me, thanks!

PR #13, opened 2019-12-20
Resource preparation

trws:
This is a collection of fixes to prepare the Resource types to be used as part of a generic allocation mechanism, and includes a helper (based heavily on what you were starting with @davidbeckingsale) for getting a resource type based on the platform value.

This should be enough to get going with to unify the cuda and hip paths of stuff.  There's some more to be done for me to consider this really "finished" though.

1. [ ] We don't have error handling or propagation on memcpy/memset in these, need a good way to handle it
2. [ ] Platform is weirding me out, we have enough to differentiate for the memory platforms for host cuda and hip, but resources need to work for TBB and openmp, and openmp offload too for that matter.  I think we may need to be able to differentiate between execution and memory space as a next-step, and would be interested in thoughts on this.

@rhornung67, @artv3, take a look here if you would.  @artv3, I think this solves all the external interface issues we talked about, let me know if I missed something.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,11 +4,11 @@ project (camp
   LANGUAGES CXX
   VERSION 0.1.0)
 
-option(ENABLE_CUDA "whether to build with cuda" Off CACHE BOOL)
-option(ENABLE_CLANG_CUDA "use clang cuda support rather than nvcc" Off CACHE BOOL)
-option(ENABLE_HIP "whether to build with hip" Off CACHE BOOL)
-option(ENABLE_TARGET_OPENMP "whether to build with OpenMP offload" Off CACHE BOOL)
-option(ENABLE_TESTS "whether to configure and build tests" On CACHE BOOL)
+option(ENABLE_CUDA "whether to build with cuda" Off)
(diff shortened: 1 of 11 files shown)


artv3:
Thanks @trws, I will give this a spin and see if we can integrate it into examples + our unit test.

davidbeckingsale:
@trws - this looks really good, thanks!

For the execution vs. memory platform, I think we probably want to have both (and our current platform is really an execution platform). We can then have a way of interrogating whether a given memory platform is valid for an execution platform, and vice versa. What do you think?

For error handling, do we want to use exceptions?

davidbeckingsale:
This is failing on Windows, see https://ci.appveyor.com/project/trws/raja/builds/30145217

trws:
The solution I just pushed up should solve the windows issue, and clean  up some other potential issues in a few places.  It's a slightly unsatisfying solution in that there needs to be better sfinae protection on get and try_get, but to build that reasonably needs the other PR that's stuck, so I'm inclined to go with the working but unsatisfying for now.

davidbeckingsale:
Excellent, thank you @trws

davidbeckingsale:
🎉

PR #17, opened 2020-02-19
Detecting OMP 4.5 for CAMP_HAVE_OMP_OFFLOAD def

mdavis36:
Some builds (gcc <= 5?) in CHAI are failing because of the 'depend' and 'nowait' clauses in omp_targets.hpp . I think those are 4.5 features? Updating to this seems to fix those builds.


trws:
Agreed, thanks @mdavis36 and @davidbeckingsale for the quick review.

PR #22, opened 2020-03-05
Adding override to Resource::get_event and wait_for.

mdavis36:
This fixes builds with clang that were failing in umpire for the resource test.
The test was failing specifically when -DENABLE_DEVELOPER_DEFAULTS=On is used.


davidbeckingsale:
The option @mdavis36 is one in Umpire that turns on pedantic warnings.

trws:
Makes sense, thanks @mdavis36.

PR #24, opened 2020-03-20
generalized camp::seq_at to work with all camp::int_seq instead of just camp::idx_seq

ajkunen:


trws:
Thanks @ajkunen, sorry this ended up sitting but it looks good!

PR #25, opened 2020-03-31
resource: fix copy construction bug

trws:
We apparently had two bugs here, both caught by @rhornung67 with a new
case we somehow hadn't hit before.

1. The internal context was sometimes being constructed with a reference
   type when it should have been the bare type
2. The rvalue-reference constructor was being used even for copy
   construction, which made Resource try and wrap an already erased
   Resource.

Both of these were preventing copying and storing a Resource by value.
This should fix both issues, and adds a minimal test to ensure we don't
regress.

@rhornung67, any chance you could test this and see if it lets you work the way you want? The intention is these should be (relatively) cheaply copyable.  We may optimize out the shared_ptr to help that at some point but it's a convenient way to start out.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -14,13 +14,15 @@ http://github.com/llnl/camp
 #include <cstring>
 #include <memory>
 #include <mutex>
+#include <type_traits>
 
-#include "camp/resource/event.hpp"
-#include "camp/resource/platform.hpp"
-#include "camp/resource/host.hpp"
+#include "camp/helpers.hpp"
 #include "camp/resource/cuda.hpp"
+#include "camp/resource/event.hpp"
 #include "camp/resource/hip.hpp"
+#include "camp/resource/host.hpp"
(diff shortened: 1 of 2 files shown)


rhornung67:
@trws Thanks for responding so quickly. I'll give this a whirl this morning and report.

rhornung67:
@trws no rush, but are you going to merge this?

trws:
Yes, I was on home schooling duty yesterday and still catching up today.

Sent from my iPhone

On Apr 2, 2020, at 12:47 PM, Rich Hornung <notifications@github.com> wrote:

﻿

@trws<https://github.com/trws> no rush, but are you going to merge this?

trws:
Tried to merge it right after the comment, but the iOS github app was completely failing on me.  At least the web version worked...

rhornung67:
Thanks Tom. This may not be ready to pull into RAJA yet (not sure why). I sent you an email to your lab address with a bit more info.

PR #28, opened 2020-04-21
Fixing unused var warning on nvcc

mdavis36:
CAMP_ALLOW_UNUSED_LOCAL was still returning unused variable warnings when compiling with cuda, this seems to hush nvcc up.

### DIFF
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -21,7 +21,11 @@ http://github.com/llnl/camp
 namespace camp
 {
 
-#define CAMP_ALLOW_UNUSED_LOCAL(X) [&X]{}()
+#if defined(__NVCC__)
+#define CAMP_ALLOW_UNUSED_LOCAL(X) sink(X)
+#else
+#define CAMP_ALLOW_UNUSED_LOCAL(X) (void)(X)
+#endif
 
 #if defined(__clang__)
 #define CAMP_COMPILER_CLANG
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -25,7 +25,6 @@ namespace camp
 #define CAMP_ALLOW_UNUSED_LOCAL(X) sink(X)
 #else
 #define CAMP_ALLOW_UNUSED_LOCAL(X) (void)(X)
-
 #endif
 
 #if defined(__clang__)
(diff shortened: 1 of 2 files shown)


include/camp/defines.hpp
```diff
@@ -21,7 +21,7 @@ For details about use and distribution, please read LICENSE and NOTICE from
 namespace camp
 {
 
-#define CAMP_ALLOW_UNUSED_LOCAL(X) (void)X
+#define CAMP_ALLOW_UNUSED_LOCAL(X) [&X]{}()
```
davidbeckingsale, line 24: Is this the new way to silence those warnings? Any pros/cons over something like `static_cast<void>(X)`? @trws?

mdavis36: I tried with static_cast however I was still getting warnings with nvcc 😞 
I just gave each version a spin on [godbolt](https://godbolt.org/z/9xod8N), using intel,gcc and clang the lambda is less performant with -O0 but -O1 or higher optimizes all solutions away the same. msvc however doesn't seem to want to optimize the lambda away quite as nicely.

mdavis36: Perhaps this might be better ?
`#if defined(__NVCC__)`
`#CAMP_ALLOW_UNUSED_LOCAL(X) [&X]{}()`
`#else `
`#CAMP_ALLOW_UNUSED_LOCAL(X) (void)X`
`#endif `

davidbeckingsale: Yea, that could be good if you were noticing some optimization differences.

trws: I'd rather avoid the lambda.  If we need something like this, how about using camp::sink, which I just realized I forgot to make the arguments const references, but either way that should pretty much always optimize out, and we can add some extra flags to it to make sure.  It's equivalent to what boost uses for their `ignore_unused` basically, like this:

```c++
template<typename...Ts>
<should be force inline> inline constexpr void sink(Ts const&...) {
}
```

mdavis36:
@trws I added the const ref and inline to sink. I'm not sure how to alias the sink function with C++11. Do you perhaps have an example of aliasing a packed template function I could look at? From what I read it doesn't seem like it's possible, but that could be wrong. For the meantime I've left it in the macro.

trws:
That's fine, there's no way to do a real alias portably, I really meant having another function with that name.  The only other thing I'd do here is add the macro to make it constexpr in c++14 mode, then LGTM.

mdavis36:
@trws Added.

trws:
As soon as this comes back green from travis I'll merge, thanks @mdavis36!

mdavis36:
Okay Travis was doing some strange things, it's good to go now 😀

PR #30, opened 2020-05-20
Default stream implementation + enable_if_t

mdavis36:
This PR is influenced by the Resource RAJA::forall PR [763](https://github.com/LLNL/RAJA/pull/763). 

It adds constructor flags for GPU resources such that a user can define if they want the resource to get the next stream in the pool or have the stream assigned to the default stream.

This PR also adds camp::enable_if_t which is the same as camp::enable_if but can return a user defined type.


include/camp/concepts.hpp
```diff
@@ -167,6 +167,9 @@ namespace concepts
   template <typename... Args>
   using enable_if = typename std::enable_if<all_of<Args...>::value, void>::type;
 
+  template <typename T, typename... Args>
```
davidbeckingsale, line 170: You might want to implement enable_if using enable_if_t now?

include/camp/resource/cuda.hpp
```diff
@@ -69,14 +73,16 @@ namespace resources
...
-      Cuda(int group = -1) : stream(get_a_stream(group)) {}
+      Cuda(int group = NEXT_STREAM) : stream(get_a_stream(group)) {}
 
       // Methods
       Platform get_platform() { return Platform::cuda; }
       static Cuda &get_default()
       {
-        static Cuda h;
+        static Cuda h(static_cast<cudaStream_t>(0));
         return h;
```
mdavis36, line 86: @trws @davidbeckingsale I've been thinking about this and perhaps using a private constructor for setting the stream on the static get_default might be better than allowing users to make multiple default stream cuda objects. What do you think?

include/camp/concepts.hpp
trws: This implementation looks good, but I'm not sure if I want to take it right now rather than reworking it.  I want to have this, but our current definition of enable_if is already a bit brain-bending based on the standard version, and this one takes that and makes it backwards... It's on my list to rectify this and various other naming and interface inconsistency issues, and it's highly related to the concepts rework material.  I guess I'm fine with it as long as we try not to spread it too far to minimize the breakage when we come back to fix this stuff.

include/camp/resource/cuda.hpp
trws: I'm not sure I quite get what you mean.

include/camp/resource/stream.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#if defined(CAMP_HAVE_CUDA) || defined(CAMP_HAVE_HIP)
+namespace camp
+{
+namespace resources
+{
+  inline namespace v1
+  {
+
+    constexpr int DEFAULT_STREAM = -1;
```
trws, line 23: I like the constexpr ints here, but I'm thinking about the values a bit.  What do you think of making `DEFAULT_STREAM == 0`, then `NEXT_STREAM` keeps its default value, and we shift lookups by one, then provide this for all devices.  I ask this because it's conventional in a large number of places for 0 to be used as the default stream in cuda code, so it might help keep things consistent.

include/camp/resource/cuda.hpp
mdavis36: So instead of having DEFAULT_STREAM and NEXT_STREAM flags we return the Cuda/Hip constructor back to what it was before these changes (removing the change on line 79 and the new stream.hpp file). This will only allow users to get a Cuda resource device with the default stream through the get_default function

trws: Oh I see.  That would help make it more like a singleton, it would make it posibly more self-documenting, but harder to write code that will use either a new stream or the default stream based on an argument.

include/camp/resource/stream.hpp
mdavis36: I think I see what you mean, but just to confirm. If a user explicitly calls Cuda(1) they are actually given the 0th stream in the pool, Cuda(0) gives a default and Cuda(-1) would return the next stream in the pool?

trws: That's what I was thinking.  It's not objectively "better" but it is more like what cuda does by default, and I'm trying to figure out if that's what we want or not...

include/camp/resource/cuda.hpp
mdavis36: Maybe we could have both? :thinking:

davidbeckingsale:
Sorry for all the noise here, I misread what the code was currently doing.

Based on our discussions this week, I think we can simplify this a bit:

- [ ] Have get_default return a build time configured default (either 0, or a regular stream)
- [ ] Have the constructor always return a stream from the pool, never the default

These changes would then mean that the usage of the default would always be explicit.

trws:
Since we discussed the defaulting on the calls, would you prefer to look at reworking this @mdavis36 or closing this one and working up a separate PR?

mdavis36:
I'll rework this branch

trws:
Works for me, I'm going to put this in draft state for now.  Trying to take today to clear down in-flight camp stuff, so don't take this so much as me poking for it as trying to clean up and figure out the status of everything.

mdavis36:
@davidbeckingsale @trws Could we get another look over this, I think this PR is ready to go.

include/camp/resource/cuda.hpp
```diff
@@ -85,15 +85,26 @@ namespace resources
...
-        return h;
+        static cudaStream_t s;
+        static std::once_flag m_onceFlag;
+        std::call_once(m_onceFlag, [] {
+#if CAMP_USE_PLATFORM_DEFAULT_STREAM
+          s = 0;
+#else
+          cudaStreamCreate(&s);
+#endif
+        });
```
MrBurmark, line 105: I think the explicit once_flag can be avoided here.
```suggestion
        static cudaStream_t s = [] {
          cudaStream_t s_ = 0;
#if CAMP_USE_PLATFORM_DEFAULT_STREAM
          cudaStreamCreate(&s_);
#endif
          return s_;
        }();
```

include/camp/resource/cuda.hpp
```diff
@@ -85,15 +85,26 @@ namespace resources
...
+        static Cuda c(s);
```
MrBurmark, line 106:
```suggestion
        static Cuda c( [] {
          cudaStream_t s = 0;
#if !CAMP_USE_PLATFORM_DEFAULT_STREAM
          cudaStreamCreate(&s); // this should be error checked
#endif
          return s;
        }());
```

trws:
@mdavis36, do you have time to rebase this? I'm going to take one more pass in a minute, but I'm pretty good with this if we can merge it.

include/camp/resource.hpp
```diff
@@ -155,6 +155,24 @@ namespace resources
...
+    EventProxy(EventProxy &&) = default;
+    EventProxy(EventProxy const &) = delete;
+    EventProxy &operator=(EventProxy &&) = default;
+    EventProxy &operator=(EventProxy const &) = delete;
+
+    EventProxy(Res* r) :
+      resource_{r}
+    {}
+
+    operator Event() const {
```
trws, line 169: Also please add an explicit `get` here.  I think having the conversion is fine, but I always prefer having an explicit option there as well.

mdavis36:
@trws I added the explicit get, do you think this is good to go

PR #32, opened 2020-06-11
Fix camp invoke

MrBurmark:
Fix camp invoke by adding decay to TupleLike when getting tuple_size


MrBurmark:
Ya, but at least types don't spontaneously decay in computer science. Code on the other hand seems to have a half life.

PR #37, opened 2020-06-29
Faster sequences

trws:
Incorporate more checks for integer sequence building intrinsics.  This makes explicit a number of checks for nvcc, icpc, xl, msvc and pgi so we get intrinsics working in as many places as possible for sequence building and pack element selection for `at<>`.

The libstdc++ and libc++ libraries use these at newer versions as well, so we could consider using these checks and deferring to the native `std::make_integer_sequence` in the future.  I'm vaguely for keeping the setup as-is, if only because it makes it easier for us to ensure we get a reasonably fast implementation everywhere.

### DIFF
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -88,10 +88,39 @@ namespace camp
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
 
-#if defined(__has_builtin)
-#if __has_builtin(__make_integer_seq)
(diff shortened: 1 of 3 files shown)


trws:
@dshollman it turns out I had failed to pull some of this into the main version of camp, but this is the set of checks and so-forth I mentioned before for getting the builtins going across compilers.  It looks like libstdc++ did add support for the clang intrinsic at some point, but the cross-section of which compiler supports what when is pretty odd.  Anyway, it would be awesome if you could find a minute to glance over this.

PR #38, opened 2020-07-09
Remove hip workarounds

trws:
The hipcc workarounds from before 3.5 break 3.5.  This gets hip working again, and improves build times on there as well, but means we *require* hip 3.5.  Upside is there is now absolutely no hip specific code in the tuple, which is kinda awesome.

@MrBurmark, @rhornung67


trws:
Excellent, thanks @davidbeckingsale and @rhornung67.

PR #39, opened 2020-07-14
Add non-defaulted tuple_helper copy constructor

MrBurmark:
This fixes an issue encountered with cuda 10.1 where the device copy constructor may not being generated in complex code.


MrBurmark:
This was addressed by @trws in https://github.com/LLNL/RAJA/pull/869.

PR #40, opened 2020-07-29
update appveyor and msvc checks

trws:


trws:
@davidbeckingsale, have a look at this.  After some testing with godbolt and windows, it looks like `__type_pack_element` just doesn't work on msvc, how it passed here I don't know, but I'm going to find out after this appveyor update comes back.  For now I'm leaving the 2017 version in, but adding the 2019 version so we can keep up to date.

PR #42, opened 2020-08-25
Add pointer rem & add helper.

rchen20:
Allows use of removal and adding of * and ** to types.

### DIFF
--- a/include/camp/helpers.hpp
+++ b/include/camp/helpers.hpp
@@ -89,10 +89,6 @@ namespace type
     struct rem_s<T*> {
       using type = T;
     };
-    template <class T>
-    struct rem_s<T**> {
-      using type = T;
-    };
 
     /// remove pointer from T
     template <class T>
@@ -102,13 +98,6 @@ namespace type
     template <class T>
     using add = T*;
   }  // end namespace ptr
-
-  namespace lvptr
-  {
-    /// add lvalue reference to T
-    template <class T>
-    using add = T**;
-  }  // end namespace lvptr
 
   namespace c
   {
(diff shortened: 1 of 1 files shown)


include/camp/helpers.hpp
```diff
@@ -79,6 +79,37 @@ namespace type
...
+  {
+    template <class T>
+    struct rem_s {
+      using type = T;
+    };
+    template <class T>
+    struct rem_s<T*> {
+      using type = T;
+    };
+    template <class T>
```
trws, line 92: This specialization needs to come out, or it will eat a double pointer down to a base type in one step, which I don't think is what we want.

include/camp/helpers.hpp
```diff
@@ -79,6 +79,37 @@ namespace type
...
+    /// remove pointer from T
+    template <class T>
+    using rem = typename rem_s<T>::type;
+
+    /// add remove pointer to T
+    template <class T>
+    using add = T*;
+  }  // end namespace ptr
+
+  namespace lvptr
```
trws, line 106: This whole namespace block and contents should come out too, the rvref is because you need two refs added to get an rvalue reference, if you want to have one for working directly with double pointers, I'd actually be fine with that, but I'd rename it to make that clear and have both add and rem variants for it.

include/camp/helpers.hpp
rchen20: I see, there could be other qualifiers in between. Done.

include/camp/helpers.hpp
rchen20: Makes sense, I removed this namespace to keep it consistent.

trws:
Excellent, thanks!

PR #45, opened 2020-12-08
Update hip.hpp to use hipMalloc

artv3:
hipMallocManaged appears to be broken with rocm 3.6.0 on LC platforms, this PR will switch to hipMalloc

### DIFF
--- a/include/camp/resource/hip.hpp
+++ b/include/camp/resource/hip.hpp
@@ -106,8 +106,10 @@ namespace resources
       template <typename T>
       T *allocate(size_t size)
       {
+        //Dec 7th 2020, hipMallocManaged is broken with HIP 3.6.0
+        //use hipMalloc instead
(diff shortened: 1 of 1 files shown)


davidbeckingsale (approved):

trws:
If this is the solve I'm good with this, but do we have anything on what the issue is with 3.6?  Is it solved in 3.7?

trws:
Does this issue exist with current hip/rocm? I'd rather keep it using the managed allocation if we can, but if we can't, this needs to be rebased to resolve conflicts so we can get it in.

davidbeckingsale:
I don't think this is necessary anymore. In Umpire we switched back to using hipMallocManaged.

davidbeckingsale:
ping @artv3

artv3:
Correct, closing PR.

PR #46, opened 2020-12-11
Making sure omp target is not compiled on TOSS, assuming no GPUs.

rchen20:
This seems to be causing some TOSS3-gcc6 failures in RAJA https://github.com/LLNL/RAJA/pull/930. GCC-6.1.0 has an advanced enough OpenMP version which turns on omp-target in CAMP. This problem correctly manifests itself as compilation errors when run manually. However, Travis compiles successfully but segfaults at runtime, which I'm not sure why.

### DIFF
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -98,7 +98,7 @@ namespace camp
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
-#endif  // !defined(__x86_64__)
+#endif
 
 // This works for:
 //   clang
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -94,9 +94,11 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
+#if !defined(__x86_64__)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
+#endif  // !defined(__x86_64__)
 
 // This works for:
 //   clang
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -94,7 +94,7 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
-#if !defined(__x86_64__)
+#if defined(ENABLE_TARGET_OPENMP)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -97,6 +97,9 @@ namespace camp
 #if defined(ENABLE_TARGET_OPENMP)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
+#else
+#define CAMP_HAVE_OMP_OFFLOAD 0
+#warning Compiler does NOT support OpenMP Target Offload even though user has enabled it!
 #endif
 #endif
 


trws:
I get the goal, but I'm not sure I'm onboard with the solution.  This seems like it will prevent issues with gcc on x86, but prevent say clang on x86 from using offload, and probably leave gcc broken on ppc64.  Is there something more specific, or better aligned, we can check?  Perhaps for GCC as compiler and a version?

rchen20:
> I get the goal, but I'm not sure I'm onboard with the solution. This seems like it will prevent issues with gcc on x86, but prevent say clang on x86 from using offload, and probably leave gcc broken on ppc64. Is there something more specific, or better aligned, we can check? Perhaps for GCC as compiler and a version?

Right, I'd like to just prevent only gcc <= 6 and only on x86 from activating this flag; I'll make this more specific. But we want other compilers to still compile offload code even if there is no GPU on the platform?

trws:
What do you mean “if there is no GPU on the platform”?  OpenMP 
target could offload to host, another node, a GPU, whatever, so I’m 
not sure there’s an architecture flag we can check for it that would 
work.  There are potentially other ways to do it, but why just x86?

rchen20:
I was thinking about our TOSS machines which are x86 but don’t have GPUs, with the exception of RZWHAMO. You’re right though, it should be working regardless whether there is a device.

include/camp/defines.hpp
```diff
@@ -94,9 +94,11 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
+#if !defined(__x86_64__)
```
rchen20, line 97:
```suggestion
#if defined(ENABLE_TARGET_OPENMP)
```

include/camp/defines.hpp
```diff
@@ -94,9 +94,11 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
+#if defined(ENABLE_TARGET_OPENMP)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
+#endif  // !defined(__x86_64__)
```
rchen20, line 101:
```suggestion
#endif
```

rchen20:
@trws Made this a bit cleaner. OpenMP Target should only be compiled if we specify the appropriate -DENABLE_TARGET_OPENMP flag in CMake.

include/camp/defines.hpp
```diff
@@ -94,9 +94,11 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
+#if defined(ENABLE_TARGET_OPENMP)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
+#endif
```
willkill07, line 101: Should enabling target offload with a compiler that does not support it produce an error or fallback?

rchen20: It will inherently produce compilation errors due to the lack of support for the advanced OpenMP offload clauses. I can add a printed warning here though.

Personally, I would prefer to see the compilation errors rather than falling back. For instance, if we fallback to standard OpenMP, the user may think they're just running a very slow OpenMP offload code, and may embark on a misguided journey in to performance debugging.

include/camp/defines.hpp
```diff
@@ -94,9 +94,11 @@ namespace camp
 #define CAMP_SUPPRESS_HD_WARN
 #endif
 
+#if defined(ENABLE_TARGET_OPENMP)
 #if _OPENMP >= 201511
 #define CAMP_HAVE_OMP_OFFLOAD 1
 #endif
```
rchen20, line 100:
```suggestion
#else
#define CAMP_HAVE_OMP_OFFLOAD 0
#warning Compiler does NOT support OpenMP Target Offload even though it is enabled!
#endif
```

trws:
This looks a great deal more reasonable. More work should be done so we actually have a config header for camp.  Much as I really, really wanted to avoid it there are things like this where we want to configure it at install time (or at least at compile time of the including TU).  If this solves the issue for you I'd be alright with it.

rchen20:
> This looks a great deal more reasonable. More work should be done so we actually have a config header for camp. Much as I really, really wanted to avoid it there are things like this where we want to configure it at install time (or at least at compile time of the including TU). If this solves the issue for you I'd be alright with it.

This certainly works for me; I'm using the same ENABLE_TARGET_OPENMP flag which is used by RAJA. I figured that it would help to keep things as consistent as possible with LLNL tools. Anyone using CAMP without RAJA would need to set this flag as well.

rchen20:
@trws Just a friendly reminder to push this into the main branch before the various imminent RAJA/Umpire/CAMP releases. If you've done so already, then thank you!

trws:
Looks good, thanks @rchen20!

PR #47, opened 2021-02-25
Add "CAMP_" prefix to "ENABLE_TESTS" option

gzagaris:
# Summary

The "ENABLE_TESTS" option is named too generically which leads to
name collisions when camp is used a sub-module within another
project. As a result, the CAMP tests will always be run with
the tests of the super-project, which may be using the same
option. Notably, BLT has also an option "ENABLE_TESTS", so
any BLT project that uses camp as a submodule will run into this.
This commit renames this option in order to give parent projects
more fine-grain control of whether to build CAMP's tests or not.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,7 @@ option(ENABLE_CUDA "whether to build with cuda" Off)
 option(ENABLE_CLANG_CUDA "use clang cuda support rather than nvcc" Off)
 option(ENABLE_HIP "whether to build with hip" Off)
 option(ENABLE_TARGET_OPENMP "whether to build with OpenMP offload" Off)
-option(ENABLE_TESTS "whether to configure and build tests" On)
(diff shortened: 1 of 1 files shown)


gzagaris:
@trws , at your convenience, could you take a quick look at this PR?

trws:
Hi @gzagaris, I'm happy to take the added prefix but would you mind adding something so that the initial value of `CAMP_ENABLE_TESTS` is On if `ENABLE_TESTS` is on? It'll keep it from being a breaking change and better match what the corresponding RAJA PR is doing IIUC.

gzagaris:
> Hi @gzagaris, I'm happy to take the added prefix but would you mind adding something so that the initial value of `CAMP_ENABLE_TESTS` is On if `ENABLE_TESTS` is on? It'll keep it from being a breaking change and better match what the corresponding RAJA PR is doing IIUC.

@trws -- Done. Let me know if that works for you. Thanks again for taking a look at this.

PR #48, opened 2021-04-13
SYCL Resource

homerdin:
Initial PR to add a SYCL resource.

### DIFF
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -92,7 +92,6 @@ namespace camp
 #define CAMP_HAVE_SYCL 1
 #define CAMP_DEVICE
 #define CAMP_HOST_DEVICE
-#pragma warn("Sycl compiler!")
 #define CAMP_SUPPRESS_HD_WARN
 
 #else
--- a/include/camp/defines.hpp
+++ b/include/camp/defines.hpp
@@ -88,6 +88,13 @@ namespace camp
 
 #define CAMP_SUPPRESS_HD_WARN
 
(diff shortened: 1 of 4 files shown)


include/camp/defines.hpp
```diff
@@ -88,6 +88,13 @@ namespace camp
 
 #define CAMP_SUPPRESS_HD_WARN
 
+#elif defined(SYCL_LANGUAGE_VERSION)
+#define CAMP_HAVE_SYCL 1
+#define CAMP_DEVICE
+#define CAMP_HOST_DEVICE
+#pragma warn("Sycl compiler!")
```
trws, line 95: Do you mean to leave this in?

trws (commented):
This looks nice and clean, everything seems to be set up the way it should be too.  Just want to know if you intend that warning to stay, I'd probably say remove it, otherwise LGTM.

include/camp/defines.hpp
homerdin: No, I'd forgotten to remove it.  I'll remove it now.

homerdin:
@trws I've updated the resource to work through/return a pointer to the sycl::queue.  Returning copies can cause issues with the memory associated with the queue.  

I've also updated the variable names, "stream" -> "qu"

trws:
Ok, this looks pretty good.  I need to figure out what's going on with appveyor though, almost looks like a cmake issue in the newer version over there.

trws:
Found the appveyor issue.  I'm going to merge this based on the results for linux, since the appveyor results are unrelated.  Thanks!

PR #52, opened 2021-04-27
add CAMP_EXPAND and for_each_arg

trws:
Simple implementations of some new helpers for camp.  I'm going to add in a bit more in this vein, but the lack of these was causing issues for some reduction work, so these are going up sooner.


mdavis36 (approved):
This worked exactly as expected in the reducer prototype :+1:

PR #54, opened 2021-04-27
add support for native events from EventProxy

### LINKED ISSUE #51: EventProxy does not support getting non-erased event
It would be convenient to be able to get a non-erased event type from EventProxy. It seems wasteful to create an erased event when type information is available.

trws:
Fixes #51

Add support for getting the native event type out of EventProxy.
This is now the default, with Event fixed to allow implicit conversion
from native event types and a new `operator Event()` overload on
EventProxy all tests pass in tree, need to see how this impacts codes.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -176,20 +176,20 @@ namespace resources
     using native_event = decltype(std::declval<Res>().get_event());
     
     EventProxy(EventProxy &&) = default;
-    EventProxy(EventProxy const &) = delete;
+    EventProxy(EventProxy &) = delete;
     EventProxy &operator=(EventProxy &&) = default;
-    EventProxy &operator=(EventProxy const &) = delete;
+    EventProxy &operator=(EventProxy &) = delete;
 
     EventProxy(Res r) :
       resource_{r}
     {}
 
-    native_event get() const {
-      return resource_->get_event();
+    native_event get() {
+      return resource_.get_event();
     }
 
-    operator native_event() const {
-      return resource_->get_event();
+    operator native_event() {
+      return resource_.get_event();
     }
 
     operator Event() {
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -181,7 +181,7 @@ namespace resources
     EventProxy &operator=(EventProxy &) = delete;
 
     EventProxy(Res r) :
-      resource_{r}
+      resource_{move(r)}
     {}
 
     native_event get() {
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -172,7 +172,9 @@ namespace resources
 #endif
 
   template<typename Res>
-  struct EventProxy {
+  struct EventProxy : ::camp::resources::detail::EventProxyBase {
+    using native_event = decltype(std::declval<Res>().get_event());
+    
     EventProxy(EventProxy &&) = default;
     EventProxy(EventProxy const &) = delete;
     EventProxy &operator=(EventProxy &&) = default;
@@ -182,8 +184,12 @@ namespace resources
       resource_{r}
     {}
 
-    Event get() {
-      return resource_.get_event_erased();
+    native_event get() const {
+      return resource_->get_event();
+    }
+
+    operator native_event() const {
+      return resource_->get_event();
     }
 
     operator Event() {
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -192,7 +192,7 @@ namespace resources
       EventProxy(EventProxy &&) = default;
       EventProxy(EventProxy &) = delete;
       EventProxy &operator=(EventProxy &&) = default;
-      EventProxy &operator=(EventProxy &) = delete;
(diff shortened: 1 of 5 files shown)


trws:
@MrBurmark, take a look at this, possibly give it a try building with something that uses events.  It should allow existing code to continue working for all expected patterns, but if we find something that breaks we can re-work.

mdavis36 (commented):
I think this looks good and should do what is expected, but will wait for @MrBurmark to give this a spin. Do we want to add some tests for this?

MrBurmark:
Sorry, I haven't gotten back to doing graph stuff again yet. This looks reasonable, I'll try to get back to this sometime next week.

trws:
@MrBurmark, did you ever get a chance to test this? I imagine we want it for the release but I haven't heard anything.  Will look into the appveyor failure but I'd like to hear if this design solves your issue first.

trws:
I *think* this is straightened out enough that we can at least work with it for now.  If you get a chance take a look @MrBurmark.

include/camp/resource.hpp
```diff
@@ -172,18 +172,24 @@ namespace resources
...
-  struct EventProxy {
+  struct EventProxy : ::camp::resources::detail::EventProxyBase {
+    using native_event = decltype(std::declval<Res>().get_event());
+    
     EventProxy(EventProxy &&) = default;
-    EventProxy(EventProxy const &) = delete;
+    EventProxy(EventProxy &) = delete;
     EventProxy &operator=(EventProxy &&) = default;
-    EventProxy &operator=(EventProxy const &) = delete;
+    EventProxy &operator=(EventProxy &) = delete;
```
MrBurmark, line 181: why are these not const&?

MrBurmark:
Looks good to me. I added a couple tests in my fork https://github.com/MrBurmark/camp/tree/trws/issue51 if you think its worth bringing those in.

trws:
Actually yeah, if you would push those in here that would be great @MrBurmark.

include/camp/resource.hpp
trws: I can try to switch them to that, but the event being pulled out is not const, and the conversion methods were not const, so there were some issues there.  It really could use a pass to ensure const-correctness, which I will try to do but I'd love another set of eyes passing over EventProxy and Event for that if you have the time.

MrBurmark:
@trws I can't push changes to the branch because its on your fork.

trws:
You should be able to now, you are now on the camp team with write permissions, and this PR is set to allow maintainers to push.

include/camp/resource.hpp
MrBurmark: I wasn't sure if you could delete the copy constructor, copy assignment operator unless it had the normal signature.

MrBurmark: I took a look over the Event related methods and didn't see anything out of place.

MrBurmark:
Thanks Tom, I pushed up the tests.

include/camp/resource.hpp
```diff
@@ -172,18 +172,24 @@ namespace resources
...
-      return resource_.get_event_erased();
+    native_event get() {
+      return resource_.get_event();
+    }
+
+    operator native_event() {
+      return resource_.get_event();
     }
 
     operator Event() {
```
MrBurmark, line 195: These conflict if Res is Resource

include/camp/resource.hpp
```diff
@@ -172,18 +172,28 @@ namespace resources
...
+
+    template <typename T = Res,
+              typename = typename std::enable_if<
+                  !std::is_same<typename std::decay<decltype(std::declval<T>().get_event())>::type,
+                                Event>::value>::type>
+    operator native_event() {
+      return resource_.get_event();
     }
 
     operator Event() {
```
MrBurmark, line 199: SFINAE to allow EventProxy<Resource> to work. @trws what do you think of this fix?

trws: I like it, may want to factor this out into something more easily identifiable but I love the result.

trws:
I'm going to add a quick tweak here, then I think we'll be good to go.

MrBurmark:
@trws That makes sense, and makes it more clear what you will get when you use an erased resource vs a non-erased resource.

include/camp/resource.hpp
MrBurmark: Ok, I am still confused about the signature change so I read the cppreference copy constructor page. I thought the copy constructor had to be T(T const&,) but I see it can also be T(T&) so deleting either signature should work to delete the copy constructor. I don't see how the EventProxy copy constructor signature matters much here especially since its deleted.

MrBurmark (approved):

include/camp/resource.hpp
trws: That makes sense, I'll try switching it back, and as you say that should actually behave just fine.  If it does it certainly makes more sense that way.

trws: Seems to behave just fine, so I reverted it.  Assuming tests pass, I'm planning to merge this version.

Issue #44, opened 2020-10-29
External CUDA streams


artv3:
Hi Folks, 

We are having trouble keeping track of streams in RAJA. To enable interplay of external software that will also be using CUDA, could there be an option for users to provide an external stream? 

Thanks,
Arturo

artv3:
The use case comes from MFEM, see PR https://github.com/mfem/mfem/pull/1852

trws:
Sure, I think we intended for that to be an option from the start, is the extra constructor for the Cuda resource not an option?

https://github.com/LLNL/camp/blob/master/include/camp/resource/cuda.hpp#L89

It occurs to me looking at it that the device number defaulting to 0 might be inappropriate there, but it should work to let you construct a resource from an arbitrary external stream.

davidbeckingsale:
That constructor is private...

trws:
Yup...  Clearly not firing on all cylinders today. 🤦 

Would making that public solve the issue for you @artv3? I see no reason we couldn't offer that kind of setup, or possibly a maker function if we privatized it to disambiguate between that and the integral version (streams just being ints on the inside).

trws:
Oh, and I think it is used in one spot, the `get_default` method constructs the resource it returns from a stream.

artv3:
Yes that should fix the problem @trws , thanks!

PR #61, opened 2021-05-12
Move private hip/cuda constructor to public

artv3:
Resolves issue #44

### DIFF
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -85,7 +85,6 @@ namespace resources
         return streams[num % 16];
       }
 
-    private:
       Cuda(cudaStream_t s, int dev=0) : stream(s), device(dev) {}
     public:
       Cuda(int group = -1, int dev=0) : stream(get_a_stream(group)), device(dev) {}
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -85,9 +85,21 @@ namespace resources
         return streams[num % 16];
       }
 
+      // Private from-stream constructor
       Cuda(cudaStream_t s, int dev=0) : stream(s), device(dev) {}
+
     public:
       Cuda(int group = -1, int dev=0) : stream(get_a_stream(group)), device(dev) {}
+
+      /// Create a resource from a custom stream
+      /// The device specified must match the stream, if none is specified the
+      /// currently selected device is used.
+      static Cuda CudaFromStream(cudaStream_t s, int dev=-1) {
+        if (dev < 0) {
+          campCudaErrchk(cudaGetDevice(&dev));
+        }
+        return Cuda(s, dev);
+      }
 
       // Methods
       Platform get_platform() { return Platform::cuda; }
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -24,6 +24,7 @@ namespace resources
 {
   inline namespace v1
   {
+    class Cuda;
 
     namespace
     {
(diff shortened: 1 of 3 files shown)


trws:
I'm good with this change, vaguely nervous we might end up with that being ambiguous in some case down the line but if it's passing now it should be alright.  Maybe put in a test for it just to make sure this doesn't slip past us?

artv3:
> I'm good with this change, vaguely nervous we might end up with that being ambiguous in some case down the line but if it's passing now it should be alright. Maybe put in a test for it just to make sure this doesn't slip past us?

Sure! Sounds good. Regarding the test, do you just mean exercise calling the constructor with a specified stream? Just want to double check what the test should look like.

trws:
Yup, pretty much just construct one, maybe do something with it and wait on it.  The main worry is that we end up with the call to the constructors being ambiguous on some platform, so anything that actually calls it should show us that.

artv3:
@trws , I pushed up a couple of test, but I couldn't figure out how to make one resource wait for another. Additionally, when trying to test on an AMD platform, I tried configuring the code via `cmake -DENABLE_HIP=ON ..' from a build directory and it wasn't picking up the hip compiler -- so the hip tests are untested right now.  If you suggestions I can make the adjustments for the tests.

trws:
Ok, the tests work.  Two last things:

1. rebase or merge so we don't have conflicts
2. remove the `dev` argument from the stream constructor, it should never have been there since the stream encodes the device already and it's just a foot-gun.

artv3:
> Ok, the tests work. Two last things:
> 
>     1. rebase or merge so we don't have conflicts
> 
>     2. remove the `dev` argument from the stream constructor, it should never have been there since the stream encodes the device already and it's just a foot-gun.

Regarding 2. If we remove the device specification that triggers the guard error: 

```
C++ exception with description "campCudaErrchk(cudaSetDevice(device)) invalid device ordinal /camp/include/camp/resource/cuda.hpp:34" thrown in the test body.
```

Reading around, I think we can get the device ID based on a given context. Perhaps something with the sequence of calls?
 
```
CUresult cuStreamGetCtx ( CUstream hStream, CUcontext* pctx );
CUresult cuCtxSetCurrent ( CUcontext ctx ) 
CUresult cuCtxGetDevice ( CUdevice* device ) 
```

Is that what you had in mind?

trws:
Yeah basically, be sure to pop the stream’s context back off though so we don’t mess up the calling code.

On May 23, 2021, at 3:49 PM, Arturo Vargas ***@***.***> wrote:

﻿

Ok, the tests work. Two last things:

1. rebase or merge so we don't have conflicts

2. remove the `dev` argument from the stream constructor, it should never have been there since the stream encodes the device already and it's just a foot-gun.


Regarding 2. If we remove the device specification that triggers the guard error:

C++ exception with description "campCudaErrchk(cudaSetDevice(device)) invalid device ordinal /camp/include/camp/resource/cuda.hpp:34" thrown in the test body.


Reading around, I think we can get the device ID based on a given context. Perhaps something with the sequence of calls?

CUresult cuStreamGetCtx ( CUstream hStream, CUcontext* pctx );
CUresult cuCtxSetCurrent ( CUcontext ctx )
CUresult cuCtxGetDevice ( CUdevice* device )


Is that what you had in mind?

artv3:
@trws, one quick question. It looks like I need to add the `cuda.h` header; however I'm not sure how to add the `-lcudart` flag to the build system. Otherwise we get the following error: 
```
undefined reference to `cuStreamGetCtx'
```

artv3:
@trws I seem to be having some trouble with CMAKE here. I tried adding cudart to the `target_link_libraries` under `test/CMakeLists.txt` like this: `target_link_libraries(${TESTNAME} camp_cuda cudart)`.  But still hit the same error... any chance you might be able to push up a fix? --- I'm feeling a bit stuck here

trws:
Have a look at this version.  I rebased it and reworked it a bit.  My main reason for wanting the argument gone was that in some cases it's possible the constructor signatures would be ambiguous, which would be pretty bad, so I left it private in case we need to do something about that and made a new static method `CudaFromStream` to create the thing. This avoids the context issue and takes care of the merge conflicts.  Work for you @artv3?

artv3:
Looks great to me! Thanks!

PR #62, opened 2021-05-17
Add error checking to cuda and hip calls

MrBurmark:
Add error checking to cuda and hip calls and throw an exception when there is a cuda error. The creation of the error string is not very elegant, suggestions?


trws:
Ready for merge?

MrBurmark:
If you're ok with it, sure.

PR #64, opened 2021-06-01
EventProxy stores resource by value.

mdavis36:
Event Proxy stores the internal resource by value. 

This is in relation to https://github.com/LLNL/RAJA/pull/1054

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -172,19 +172,19 @@ namespace resources
     EventProxy &operator=(EventProxy &&) = default;
     EventProxy &operator=(EventProxy const &) = delete;
 
-    EventProxy(Res* r) :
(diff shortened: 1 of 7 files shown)


trws:
I like this direction.  Assuming it's tweaked to get the tests green LGTM.

MrBurmark:
Should resource get_default now return by value as well?

trws:
For encapsulation reasons I suppose that might make sense.  It wouldn't really change anything for dangling references unless someone uses `decltype(auto)` or similar, but it wouldn't hurt either and would make it impossible for callers to modify the default value through it which seems like a good thing.

mdavis36:
@trws @MrBurmark is this good to go?

trws (approved):
I'm happy. :shipit:

PR #65, opened 2021-06-08
Hip resource device guards.

mdavis36:
This PR adds device and memory size guards to mirror the functionality of Cuda resource. The size guards on memory operations are needed to get some failing tests to work in RAJA.

### DIFF
--- a/include/camp/resource/hip.hpp
+++ b/include/camp/resource/hip.hpp
@@ -24,6 +24,21 @@ namespace resources
   inline namespace v1
   {
 
+    namespace
+    {
+      struct device_guard {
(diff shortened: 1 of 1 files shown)


mdavis36:
@trws This is green now, are you good with this? we need it for https://github.com/LLNL/RAJA/pull/1054

trws:
Are the formatting changes from clang-format using the project config or something else? Otherwise I think this is fine, would prefer to have a test that at least calls it though.

mdavis36:
I just copied the formatting from the cuda resource file.

trws:
Fair enough.

PR #66, opened 2021-06-16
add get_event_erased to resource class

artv3:
Since we pass back a erased resource type in https://github.com/LLNL/RAJA/pull/1029, we need to add get_event_erased method class to the resource class.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -91,7 +91,7 @@ namespace resources
         m_value->memset(p, val, size);
       }
       Event get_event() { return m_value->get_event(); }
-      Event get_event_erased() { return m_value->get_event(); }
+      Event get_event_erased() { return m_value->get_event_erased(); }
       void wait_for(Event *e) { m_value->wait_for(e); }
 
     private:
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -91,6 +91,7 @@ namespace resources
         m_value->memset(p, val, size);
       }
       Event get_event() { return m_value->get_event(); }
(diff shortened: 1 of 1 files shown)


include/camp/resource.hpp
```diff
@@ -91,6 +91,7 @@ namespace resources
         m_value->memset(p, val, size);
       }
       Event get_event() { return m_value->get_event(); }
+      Event get_event_erased() { return m_value->get_event(); }
```
trws, line 94:
```suggestion
      Event get_event_erased() { return m_value->get_event_erased(); }
```

trws (changes requested):
With the suggested change, or an equivalent one, I'm good with this.

trws:
@artv3: Error from Travis, looks like the ContextInterface doesn't have the method to pass it through:

```
1: /home/raja/include/camp/resource.hpp:94:50: error: 'class camp::resources::v1::Resource::ContextInterface' has no member named 'get_event_erased'
```

artv3:
Thanks @trws, could we sync up after the release to try to fix this? I'm off the next few days but hopefully after we can sort it out. 
This PR is needed for the following RAJA branch - https://github.com/LLNL/RAJA/pull/1029.

trws:
Sure.  The problem should be solved by adding a `get_event_erased` right next to the `get_event` proxy method [here](https://github.com/LLNL/camp/pull/66/files#diff-a31e2568f20cbb0cdffc68f85d78c21e6df4a6ac05561a1a0205befd4336554cL106).

artv3:
Hooray! all test pass!

trws (approved):
Looks great, thanks!

PR #73, opened 2021-08-03
add memory access type to allocation routines

trws:
@rhornung67

Here's my first cut at adding different allocation types to the resources.  I decided to start it off as an argument that's defaulted to device, but if we want them to be separate functions that's an option.  There are three modes, device, pinned, and managed currently.  I'm tempted to make it device, *shared* and managed so that it's based on the access semantics of the memory rather than the location, but most people aren't used to that (it's what we've been working on for OpenMP, pinned and true shared can provide the same stronger guarantee than managed can, but on cuda, sycl and hip both shared and managed are accessed through the same function).

One aspect I'm not terribly fond of is having to pass the access type to the free as well.  While SYCL doesn't need it, and OpenMP probably won't need it when it matters, technically both cuda and hip do.  It's possible to query the type of memory from them to decide, but it's not a fast operation, and they don't do it internally, so we have to either always incur that cost, or pass the requirement to identify the memory type to the user.

Thoughts on this would be appreciated.

### DIFF
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -200,7 +200,7 @@ namespace resources
       }
       void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device)
       {
-        void *p = allocate<char>(size);
+        void *p = allocate<char>(size, ma);
         this->memset(p, 0, size);
         return p;
       }
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -37,8 +37,8 @@ namespace resources
 
         ~device_guard() { campCudaErrchk(cudaSetDevice(prev_device)); }
 
-      int prev_device;
-    };
+        int prev_device;
+      };
 
     }  // namespace
 
@@ -47,9 +47,12 @@ namespace resources
     public:
       CudaEvent(cudaStream_t stream) { init(stream); }
 
-      CudaEvent(Cuda& res);
+      CudaEvent(Cuda &res);
 
-      bool check() const { return (campCudaErrchk(cudaEventQuery(m_event)) == cudaSuccess); }
+      bool check() const
+      {
+        return (campCudaErrchk(cudaEventQuery(m_event)) == cudaSuccess);
+      }
       void wait() const { campCudaErrchk(cudaEventSynchronize(m_event)); }
       cudaEvent_t getCudaEvent_t() const { return m_event; }
 
@@ -58,7 +61,8 @@ namespace resources
 
       void init(cudaStream_t stream)
       {
-        campCudaErrchk(cudaEventCreateWithFlags(&m_event, cudaEventDisableTiming));
+        campCudaErrchk(
+            cudaEventCreateWithFlags(&m_event, cudaEventDisableTiming));
         campCudaErrchk(cudaEventRecord(m_event, stream));
       }
     };
@@ -92,15 +96,19 @@ namespace resources
       }
 
       // Private from-stream constructor
-      Cuda(cudaStream_t s, int dev=0) : stream(s), device(dev) {}
+      Cuda(cudaStream_t s, int dev = 0) : stream(s), device(dev) {}
 
     public:
-      Cuda(int group = -1, int dev=0) : stream(get_a_stream(group)), device(dev) {}
+      Cuda(int group = -1, int dev = 0)
+          : stream(get_a_stream(group)), device(dev)
+      {
+      }
 
       /// Create a resource from a custom stream
       /// The device specified must match the stream, if none is specified the
       /// currently selected device is used.
-      static Cuda CudaFromStream(cudaStream_t s, int dev=-1) {
+      static Cuda CudaFromStream(cudaStream_t s, int dev = -1)
+      {
         if (dev < 0) {
           campCudaErrchk(cudaGetDevice(&dev));
         }
@@ -111,7 +119,7 @@ namespace resources
       Platform get_platform() { return Platform::cuda; }
       static Cuda get_default()
       {
-        static Cuda c( [] {
+        static Cuda c([] {
           cudaStream_t s;
 #if CAMP_USE_PLATFORM_DEFAULT_STREAM
           s = 0;
@@ -123,15 +131,9 @@ namespace resources
         return c;
       }
 
-      CudaEvent get_event()
-      {
-        return CudaEvent(*this);
-      }
+      CudaEvent get_event() { return CudaEvent(*this); }
 
-      Event get_event_erased()
-      {
-        return Event{CudaEvent(*this)};
-      }
+      Event get_event_erased() { return Event{CudaEvent(*this)}; }
 
       void wait()
       {
@@ -154,31 +156,57 @@ namespace resources
 
       // Memory
       template <typename T>
-      T *allocate(size_t size)
+      T *allocate(size_t size, MemoryAccess ma = MemoryAccess::Device)
       {
         T *ret = nullptr;
         if (size > 0) {
           auto d{device_guard(device)};
-          campCudaErrchk(cudaMallocManaged(&ret, sizeof(T) * size));
+          switch (ma) {
+            case MemoryAccess::Device:
+              campCudaErrchk(cudaMalloc(&ret, sizeof(T) * size));
+              break;
+            case MemoryAccess::Pinned:
+              // TODO: do a test here for whether managed is *actually* shared
+              // so we can use the better performing memory
+              campCudaErrchk(cudaMallocHost(&ret, sizeof(T) * size));
+              break;
+            case MemoryAccess::Managed:
+              campCudaErrchk(cudaMallocManaged(&ret, sizeof(T) * size));
+              break;
+          }
         }
         return ret;
       }
-      void *calloc(size_t size)
+      void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device)
       {
         void *p = allocate<char>(size);
         this->memset(p, 0, size);
         return p;
       }
-      void deallocate(void *p)
+      void deallocate(void *p, MemoryAccess ma = MemoryAccess::Device)
       {
         auto d{device_guard(device)};
+        switch (ma) {
+          case MemoryAccess::Device:
+            campCudaErrchk(cudaFree(p));
+            break;
+          case MemoryAccess::Pinned:
+            // TODO: do a test here for whether managed is *actually* shared
+            // so we can use the better performing memory
+            campCudaErrchk(cudaFreeHost(p));
+            break;
+          case MemoryAccess::Managed:
+            campCudaErrchk(cudaFree(p));
+            break;
+        }
         campCudaErrchk(cudaFree(p));
       }
       void memcpy(void *dst, const void *src, size_t size)
       {
         if (size > 0) {
           auto d{device_guard(device)};
-          campCudaErrchk(cudaMemcpyAsync(dst, src, size, cudaMemcpyDefault, stream));
+          campCudaErrchk(
+              cudaMemcpyAsync(dst, src, size, cudaMemcpyDefault, stream));
         }
       }
       void memset(void *p, int val, size_t size)
@@ -197,7 +225,7 @@ namespace resources
       int device;
     };
 
-    inline CudaEvent::CudaEvent(Cuda& res)
+    inline CudaEvent::CudaEvent(Cuda &res)
     {
       auto d{device_guard(res.get_device())};
       init(res.get_stream());
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -16,7 +16,10 @@ http://github.com/llnl/camp
 #include "camp/resource/platform.hpp"
 
 #ifdef CAMP_HAVE_CUDA
+
 #include <cuda_runtime.h>
+
+#include <exception>
 
 namespace camp
 {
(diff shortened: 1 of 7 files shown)


davidbeckingsale:
I think this looks good - only thing that stuck out to me was that in some places you have used both shared and pinned in the actual implementation. I'm fine with either naming but I think shared is so overloaded it's potentially more trouble than it's worth.

trws:
I meant to remove the usage of shared, so that's a bug.  I'll pull that out, and thanks for finding it!

include/camp/resource/cuda.hpp
```diff
@@ -154,31 +156,57 @@ namespace resources
...
-          campCudaErrchk(cudaMallocManaged(&ret, sizeof(T) * size));
+          switch (ma) {
+            case MemoryAccess::Device:
+              campCudaErrchk(cudaMalloc(&ret, sizeof(T) * size));
+              break;
+            case MemoryAccess::Pinned:
+              // TODO: do a test here for whether managed is *actually* shared
+              // so we can use the better performing memory
+              campCudaErrchk(cudaMallocHost(&ret, sizeof(T) * size));
+              break;
```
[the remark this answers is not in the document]
trws: I had originally named it "Shared".  Reason being that while there are three kinds of memory you can allocate: pinned, truly shared, and managed single-access memory like on pascal and earlier, cuda and hip only allow allocating pinned and one of the other two.  Pinned and truly shared provide the same access guarantees, managed does not, so I was tempted to roll pinned and truly shared into one thing, and have managed be the other one.

That would be different from cuda and hip though.

trws:
I fixed the usage of `Shared` in the hip resource.

This version should be functional, but I'm not completely sure I like the interface.  What do you think of having to tell deallocate what kind of memory is being freed? I'm pretty sure it can be avoided, but it would have a performance cost to do it. I could default it to a "Detect" value or similar and do the easier but slower thing, but leave the option there to skip the cost if the user has the type handy?

rhornung67:
ooooffff. I hadn't thought about that. I think it may be best for the default be to not require the kind of memory to be specified when de-allocating, but allow it as an option. Then, make it clear in the documentation what the performance ramifications are. We can easily track this in the RAJA list segment implementation for hat use case.

rhornung67:
@trws how close is this to being merged? I think device, shared, and managed sounds reasonable, with device being the default. 

For the deallocation part, does it make sense to have the resource track which allocation mode is used to allocate, which would be set when the first resource allocation is made, and subsequent allocation with the same resource using a different mode would be a runtime error? Then, users would have to be sure to deallocate using the same resource they used for allocation.  Or is this too restrictive/clumsy?

include/camp/resource/cuda.hpp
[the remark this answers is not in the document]
MrBurmark: That's true, I believe even truly shared memory can't be used with gpu non-aware MPI on sierra though.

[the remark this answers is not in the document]
trws: It should work, but very slowly. If it actually flat doesn't work that would be a potentially serious problem, since it would mean their access guarantees aren't being met.

trws:
I kinda like the idea, but I think having it be something they can't switch might be a problem.  Perhaps it would be worthwhile to have a way to get a sub-resource with a different value, and either way it would set what the default is but allow it to be overridden?

I should be able to put up the auto-detection stuff today, if we want to be fancier with the interface in terms of having a tracking default that might take longer to figure out how to spell it.

trws:
Just pushed a version that should do the auto-detection for free.  I'm getting onto some systems to set up some tests for it, but assuming that goes alright this is good to go.

include/camp/resource/cuda.hpp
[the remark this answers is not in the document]
MrBurmark: My recollection was that it gives the wrong answer.

rhornung67:
@trws I pulled this branch into RAJA develop to test. It works fine on CPU builds (all tests compile and pass). However, trying a cuda build on blueos, I'm seeing the following compilation errors:

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(116): error #20: identifier "runtime_error" is undefined

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(116): error #20: identifier "runtime_error" is undefined

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(208): error #135: enum "camp::resources::v1::MemoryAccess" has no member "Detect"

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(208): error #135: enum "camp::resources::v1::MemoryAccess" has no member "Detect"

[  2%] Built target blt_cuda_version_smoke
/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(116): error #20: identifier "runtime_error" is undefined

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(208): error #135: enum "camp::resources::v1::MemoryAccess" has no member "Detect"

/g/g16/hornung1/AASD/RAJA-repo/raja-DEV/RAJA/tpl/camp/include/camp/resource/cuda.hpp(182): error #135: enum "camp::resources::v1::MemoryAccess" has no member "Detect"

3 errors detected in the compilation of "/var/tmp/hornung1/tmpxft_00015ad1_00000000-6_PluginStrategy.cpp1.ii".

More of the same after this.  Have you tried to build RAJA with the version of camp?

rhornung67:
Building RAJA with Hip, I see syntax errors, etc....

/g/g16/hornung1/AASD/RAJA-repo/raja-TEST/RAJA/tpl/camp/include/camp/resource/hip.hpp:100:9: error: unknown type name 'hipPointerAttributes'; did you mean 'hipPointerAttribute_t'?
        hipPointerAttributes a;
        ^~~~~~~~~~~~~~~~~~~~
        hipPointerAttribute_t
/opt/rocm-4.1.0/hip/include/hip/hip_runtime_api.h:177:3: note: 'hipPointerAttribute_t' declared here
} hipPointerAttribute_t;

....

/g/g16/hornung1/AASD/RAJA-repo/raja-TEST/RAJA/tpl/camp/include/camp/resource/hip.hpp:103:21: error: no member named 'type' in 'hipPointerAttribute_t'
          switch (a.type) {
                  ~ ^
/g/g16/hornung1/AASD/RAJA-repo/raja-TEST/RAJA/tpl/camp/include/camp/resource/hip.hpp:108:18: error: use of undeclared identifier 'hipMemoryTypeManaged'; did you mean 'hipMemoryTypeUnified'?
            case hipMemoryTypeManaged:
                 ^~~~~~~~~~~~~~~~~~~~
                 hipMemoryTypeUnified
/opt/rocm-4.1.0/hip/include/hip/hip_runtime_api.h:162:5: note: 'hipMemoryTypeUnified' declared here
    hipMemoryTypeUnified  ///< Not used currently

...

rhornung67:
I played around with the camp code and got RAJA to build with nvcc. There were also some warnings about missing cases in switch statements that I put enough stub code in to quiet those. However, all RAJA CUDA tests failed for various reasons. Apparently, this still needs some work.

include/camp/resource/hip.hpp
```diff
@@ -177,15 +181,15 @@ namespace resources
           switch (ma) {
             case MemoryAccess::Unknown:
```
rhornung67, line 182: Did you remove MemoryAccess::Detect?

trws: I changed the name so it made more sense when used as both "detect" and "I can't figure out what this is." Unknown is detect, and will auto-detect how to free the thing when passed to deallocate.

trws:
With the last two commits, hip tests pass with raja no problem on corona.  I'm still trying to track down an issue with cuda though.

rhornung67:
Great. The issues I saw earlier were not serious. It looked like the code wasn't complete and ready to be tested. If you want me to run some tests also, please let me know.

trws:
If you get a chance to do hip tests or sycl tests especially that would 
be great.  There’s a segfault in some cuda tests I want to get a 
closer look at.

trws:
I'm now wondering very much why this didn't break on hip, but the issue in the cuda tests is stuff like this:

https://github.com/LLNL/RAJA/blob/65fc72c38f720de4df0681c17a39bb1302e8f771/test/functional/forall/atomic-ref/tests/test-forall-AtomicRefAdd.hpp#L96

The hit array is allocated with a camp resource, then used in forall regions *and* used directly from the host.  Wherever this happens, having the default be `Device` will break us.  It seems like the changes are working the way they should, but some of the tests are going to be dangerous like this.

I'm vaguely tempted to see if either CHAI, or a minimal chai-like mechanism, could be added into this since the allocators are there, but that's not going to be fast.  If we want to do this and have tests pass for current release, I need to make the default "Managed" for now, otherwise we can hold back on this and let it bake for a cycle.

rhornung67:
@trws this issue is just with the atomic tests, correct? If so, @rchen20 fixing the memory accesses in these tests should be any easy fix, right? Do you have time to do that in the next day or two?

rchen20:
> @trws this issue is just with the atomic tests, correct? If so, @rchen20 fixing the memory accesses in these tests should be any easy fix, right? Do you have time to do that in the next day or two?

That should be a pretty easy fix. I'll go through the atomic tests now and make sure there are proper host and device arrays, with the correct memcpy's.

rhornung67:
@rchen20 thanks! make sure you pull in Tom's camp PR branch to test.

trws:
I'm not sure it is, there were a good number of tests with the problem, but it could be.  If so that would be fantastic!

rhornung67:
I'm pretty sure all of the tests converted to the new organization follow the pattern of using a camp resource, which does the allocation on host/device, initialization on host, copy to device, copy back to host to check results. The remaining tests that have not been converted, which will be done soon, do not use camp resources, so I think we're good there. Same with our examples and exercises.

rchen20:
I've added usage of the host resource with memcpy's with @trws's branch, but the AtomicRefAdd test is seg faulting in the first forall. Also changed the lambda decoration to `RAJA_DEVICE` just in case, and tried allocating with `camp::resources::MemoryAccess::Managed` but that produces the same seg fault. There might be something else wrong apart from the test's memory management.

rhornung67:
@rchen20 you need a host resource and a work resource to allocate data on the host and device, respectively. Here's an example: 
https://github.com/LLNL/RAJA/blob/develop/test/functional/forall/segment/tests/test-forall-ListSegment.hpp#L36

The allocateForallTestData routine is here: 
https://github.com/LLNL/RAJA/blob/develop/test/include/RAJA_test-forall-data.hpp#L18

include/camp/resource/cuda.hpp
```diff
@@ -154,31 +176,63 @@ namespace resources
...
+              campCudaErrchk(cudaMallocManaged(&ret, sizeof(T) * size));
+              break;
+          }
         }
         return ret;
       }
-      void *calloc(size_t size)
+      void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device)
       {
         void *p = allocate<char>(size);
```
rchen20, line 203:
```suggestion
        void *p = allocate<char>(size, ma);
```

rchen20: This needs to forward/pass the `MemoryAccess` type to the actual `allocate()` call.

rchen20:
The problem with the AtomicRefAdd test is in the CountOp object: https://github.com/LLNL/RAJA/blob/65fc72c38f720de4df0681c17a39bb1302e8f771/test/functional/forall/atomic-ref/tests/test-forall-AtomicRefAdd.hpp#L75
CountOp is allocated on the host, and expects to be accessed on the device as well. The test works if I manually call CAMP allocate with Pinned or Managed, but to get Device working I'll need to refactor this CountOp functionality.

rhornung67:
How involved is refactoring that test? Can the test be simplified?  To get things working soon, use pinned/managed allocations as needed. Then, make a GitHub issue describing what needs to be revisited.

include/camp/resource/cuda.hpp
trws: Thanks @rchen20, can't believe I missed that. 🤦

MrBurmark:
@rchen20 I'm not sure what you mean, the things allocated here are the count, list, and hit pointers. The CountOp object is copied as a member of the lambda.

rchen20:
I was able to refactor AtomicRefAdd with fewer major changes than I thought (thanks @MrBurmark!). I'll take a look at the other RAJA test failures throughout this week.

rchen20:
The RAJA tests are fixed/worked around here https://github.com/LLNL/RAJA/pull/1102. @MrBurmark @trws @rhornung67  Please take a look!

rhornung67:
@rchen20 this branch should be pulled in to your RAJA PR branch, correct?

rchen20:
> @rchen20 this branch should be pulled in to your RAJA PR branch, correct?

Yes, the RAJA PR requires this CAMP fork & branch.

rhornung67:
@trws can we get this PR merged soonish so we can pull these changes into the RAJA submodule and finish up the release? Thanks.

rhornung67:
Excellent. Thanks @trws for the merge and the camp release.

PR #74, opened 2021-08-04
Only change device if necessary in device guards

MrBurmark:
Do not change device if the requested device is already in use.
This let's us avoid overheads especially during profiling as these are
used frequently.


MrBurmark:
@trws Are you ok with this?

trws:
Yup, works for me. 👍

PR #77, opened 2021-08-16
Fix for MemoryAccess in SYCL resource.

rchen20:


rchen20:
@trws @rhornung67 This ought to fix the compilation error in the SYCL PR. The MemoryAccess variable was just missing from some function signatures.

rhornung67 (approved):

rchen20:
Just made another change to fix the error regarding conversion from `void *` to `char*`. @trws should edit this fix if there is something wrong with using the SYCL templated mallocs.

trws:
There shouldn't be, this looks really good, thanks for the fix @rchen20!

PR #87, opened 2021-09-14
change offload deps to use the _runtime variants

trws:
Use the runtime variants since we don't actually use compiler features in camp itself for these things.


davidbeckingsale (approved):

trws:
Thanks @davidbeckingsale, I'll merge this and post another point release in a moment.

trws:
Actually... This breaks enough of the build process and usage I'm thinking this should be a minor rather than a point release, either way it will be up shortly.

PR #92, opened 2021-12-05
ensure defines.hpp is installed properly

trws:


trws:
Testing and review off-platform by McFadden.

PR #94, opened 2021-12-07
revert get calls to old style

trws:


trws:
Tested and reviewed by McFadden off-platform.

PR #96, opened 2022-01-04
bump version to 0.4.0

trws:


CMakeLists.txt
```diff
@@ -118,7 +118,7 @@ include(CMakePackageConfigHelpers)
 
 write_basic_package_version_file(
   "${PROJECT_BINARY_DIR}/campConfigVersion.cmake"
-  VERSION 0.2.3
+  VERSION ${PACKAGE_VERSION}
```
davidbeckingsale, line 121: I think this should be `$PROJECT_VERSION` (although if the VERSION argument is not used, then PROJECT_VERSION is used by default afaik)

trws: Thanks for that, you're right.

trws:
```suggestion
  VERSION ${PROJECT_VERSION}
```

trws:
@davidbeckingsale, this should be ready for another pass when you have a minute.

davidbeckingsale (approved):

PR #98, opened 2022-02-09
Make tuple base public.

rchen20:
While the `test-kernel-single-loop-ForICount-8-OpenMPTarget` test in RAJA (https://github.com/LLNL/RAJA/pull/1207), there was a compilation error in CAMP tuple. I'm not sure if this is the right solution, but making the base tuple_helper class publicly accessible solved the build error (although the test still failed with XLC).

Note: This solution does not solve the clang internal error https://github.com/LLNL/RAJA/issues/1216.

Error:
In file included from /usr/WS1/chen59/allraja/rajaatomicexhaustive/raja_git_atomicexhaustive/build_lc_blueos-xl_omptarget-2021.12.22/test/functional/kernel/single-loop-tile-icount-tcount/test-kernel-single-loop-ForICount-8-OpenMPTarget.cpp:11:
In file included from /usr/WS1/chen59/allraja/rajaatomicexhaustive/raja_git_atomicexhaustive/test/include/RAJA_test-base.hpp:15:
In file included from /usr/WS1/chen59/allraja/rajaatomicexhaustive/raja_git_atomicexhaustive/include/RAJA/RAJA.hpp:33:
In file included from /usr/WS1/chen59/allraja/rajaatomicexhaustive/raja_git_atomicexhaustive/include/RAJA/util/camp_aliases.hpp:31:
/usr/WS1/chen59/allraja/rajaatomicexhaustive/raja_git_atomicexhaustive/tpl/camp/include/camp/tuple.hpp:154:60: error: 'base' is a private member of 'camp::tuple<RAJA::TypedRangeSegment<unsigned long long, long long> >'
  return static_cast<tpl_get_store<Tuple, index> const&>(t.base).get_inner();
                                                           ^

### DIFF
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -263,7 +263,7 @@ public:
       camp::list<Elements...>,
       camp::make_idx_seq_t<sizeof...(Elements)>>::type;
   using type = tuple;
-  Base base;
+  Base base;  // Place this back into private when XLC can handle this better. 
 
 private:
 
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -263,9 +263,9 @@ public:
       camp::list<Elements...>,
       camp::make_idx_seq_t<sizeof...(Elements)>>::type;
   using type = tuple;
+  Base base;
 
 private:
-  Base base;
 
   template <camp::idx_t index, class Tuple>
   CAMP_HOST_DEVICE constexpr friend auto& get(Tuple& t) noexcept;


rchen20:
Tagging @trws @rhornung67 because I cannot add reviewers.

trws:
Ok... I can see why that would fix that error, but that error is very, very wrong.  The `get` function is a friend function of tuple, so that should all be kosher.  Would you be willing to try making the friend functions inline into the class and see if that also works around it rather than exposing the base?

I know we haven't really done a great deal to preserve API or ABI, but that one might be nasty to fix later if it goes public now.

rchen20:
@trws Looks like Kristi is running into this same issue on the RAJA CI with XL. https://lc.llnl.gov/gitlab/radiuss/RAJA/-/jobs/561894

trws:
Ok, we need this for the release, but I really do not want to leave it this way. Would you add a TODO and an issue to come back and fix this?

include/camp/tuple.hpp
```diff
@@ -263,9 +263,9 @@ struct tuple {
       camp::list<Elements...>,
       camp::make_idx_seq_t<sizeof...(Elements)>>::type;
   using type = tuple;
+  Base base;
```
rchen20, line 266:
```suggestion
  Base base;  // Place this back into private when XLC can handle this better. 
```

rchen20:
The clang docker instance is not starting up properly. Apart from that, this should be ready to go. I added a comment to the code to remind us to change this, and the issue is up as well.

trws:
The clang thing is an upstream issue, I have a fix for it in another live PR, but this is cleared by the other clang.  Thanks @rchen20.

PR #105, opened 2022-03-28
Add apply function for tuple

adayton1:


trws:
This looks good, but how does it differ from `invoke`?  Actually glancing at it, the existing invoke may be missing a forward where one would be appropriate, but perhaps that should be rectified and `apply` be an alias to `invoke`?

trws:
Sounds good to me.  I think when we re-work the API for RAJA and CAMP on the way to 1.0 this is something to fix too, make apply the one for this and use invoke for its now established purpose.

trws (approved):

PR #106, opened 2022-03-31
Support CTAD and structured bindings with tuple

adayton1:
These are features of c++17 that make using tuples much easier.

Class template argument deduction allows the following shorthand:
`camp::tuple t{1, 2.0};`
Contribution from Sam Mish (@samuelpmishLLNL)

Structured binding also allows some shorthand:
`auto [a, b] = t;`
Contribution from Jon Wong (@jonwong12)

### DIFF
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -372,9 +372,6 @@ public:
 /// e.g. camp::tuple t{1, 2.0};
 template <class... T>
 tuple(T...) -> tuple<T...>;
-
-template <class... T>
-tagged_tuple(T...) -> tagged_tuple<T...>;
 #endif
 
 template <typename... Tags, typename... Args>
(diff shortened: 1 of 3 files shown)


adayton1:
I know these are c++17 features, but I guarded them so that it should still compile with c++14. It also sounds like RAJA/CHAI/Umpire/CAMP all want to move to c++17 anyway.

include/camp/tuple.hpp
```diff
@@ -359,14 +359,21 @@ class tagged_tuple : public tuple<Elements...>
...
   using TList = camp::list<>;
   using TMap = TList;
   using type = tuple;
 };
 
+#if defined(__cplusplus) && __cplusplus >= 201703L
+/// Class template argument deduction rule for tuples
+/// e.g. camp::tuple t{1, 2.0};
+template <typename... T>
+tuple(T...) -> tuple<T...>;
```
trws, line 374: Mind adding one for tagged tuple while you're in here? Glad to have this done, thanks for putting it up!

adayton1: Sure! Done!

adayton1: Actually, I still need to add tests.

trws: Cool, send me a ping when you're happy with the state and I'll take another look, so far this is all good with me.

adayton1: I'm not actually sure if we can provide CTAD for tagged_tuple since there's no constructor that takes the TagList. And structured binding is a little weird since the TagList also has to be pulled out.

adayton1: Oh, but if the TagList is also pulled out, there's an out of range assert.

adayton1: OK, I got structured bindings to work with tagged_tuple. I'm still attempting to figure out the CTAD.

adayton1: @trws, it looks like partial CTAD is what we would need and that is not supported in the standard. Either that, or we need a constructor that takes the TagList as well. Thoughts?

adayton1:
@trws, this is ready for your review again.

trws (approved):
I'm happy, thanks a lot for this @adayton1, good stuff.  I'll give some thought to the tagged_tuple option, but as you say there's no really obvious way to get the taglist in there, so lets go with this.

PR #109, opened 2022-05-16
Fix print_tuple namespace.

rchen20:


trws (approved):
Excellent, thanks @rchen20!

PR #119, opened 2022-10-06
Avoid naming array in CAMP_EXPAND

MrBurmark:
Avoid naming array in CAMP_EXPAND so it is more compatible with existing code. This should avoid causing errors is EXPAND is used multiple times in the same scope.
@mcfadden8 found that the previous change broke umpire as umpire passed the result to a sink like function to quiet the warning. This change will still break umpire as sink returns void, but its fixed here https://github.com/LLNL/Umpire/pull/786#pullrequestreview-1133512641.

### DIFF
--- a/include/camp/helpers.hpp
+++ b/include/camp/helpers.hpp
@@ -41,9 +41,8 @@ namespace detail
 {
   using __expand_array_type = int[];
 }
-#define CAMP_EXPAND(...) \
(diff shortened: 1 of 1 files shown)


trws (approved):

trws:
I cannot wait to be able to make this macro use C++17, thanks again @MrBurmark.

PR #125, opened 2023-01-24
Use current GPU instead of device 0

MrBurmark:
Use whatever is the current device in hip and cuda resources instead of using device 0.
Fix race conditions in the Cuda, Hip, and Sycl resource get stream and queue methods.

This is aimed at fixing this umpire issue https://github.com/LLNL/Umpire/issues/798 from @gzagaris.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ endif()
 
 if (COMPILER_FAMILY_IS_MSVC)
   if (NOT BUILD_SHARED_LIBS)
-    target_compile_definitions(camp PUBLIC CAMP_WIN_STATIC_BUILD)
+    target_compile_definitions(camp PRIVATE CAMP_WIN_STATIC_BUILD)
   else (NOT BUILD_SHARED_LIBS)
     target_compile_definitions(camp PRIVATE CAMP_DLL_EXPORTS)
   endif (NOT BUILD_SHARED_LIBS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ endif()
 
 if (COMPILER_FAMILY_IS_MSVC)
   if (NOT BUILD_SHARED_LIBS)
-    target_compile_definitions(camp PUBLIC WIN_STATIC_BUILD)
+    target_compile_definitions(camp PUBLIC CAMP_WIN_STATIC_BUILD)
   else (NOT BUILD_SHARED_LIBS)
     target_compile_definitions(camp PRIVATE CAMP_DLL_EXPORTS)
   endif (NOT BUILD_SHARED_LIBS)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ endif()
 
 if (COMPILER_FAMILY_IS_MSVC)
   if (NOT BUILD_SHARED_LIBS)
-    target_compile_definitions(camp PRIVATE CAMP_WIN_STATIC_BUILD)
+    target_compile_definitions(camp PUBLIC CAMP_WIN_STATIC_BUILD)
   else (NOT BUILD_SHARED_LIBS)
     target_compile_definitions(camp PRIVATE CAMP_DLL_EXPORTS)
   endif (NOT BUILD_SHARED_LIBS)
--- a/include/camp/config.in.hpp
+++ b/include/camp/config.in.hpp
@@ -5,6 +5,8 @@
 #cmakedefine CAMP_ENABLE_CUDA
 #cmakedefine CAMP_ENABLE_HIP
 #cmakedefine CAMP_ENABLE_SYCL
+#cmakedefine CAMP_WIN_STATIC_BUILD
+#cmakedefine CAMP_DLL_EXPORTS
 #endif
 
 #define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR@
--- a/include/camp/config.in.hpp
+++ b/include/camp/config.in.hpp
@@ -6,7 +6,6 @@
 #cmakedefine CAMP_ENABLE_HIP
 #cmakedefine CAMP_ENABLE_SYCL
 #cmakedefine CAMP_WIN_STATIC_BUILD
-#cmakedefine CAMP_DLL_EXPORTS
 #endif
 
 #define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR@
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -167,7 +167,7 @@ namespace resources
       }
 
     public:
-      Cuda(int group = -1, int dev = get_current_device())
+      explicit Cuda(int group = -1, int dev = get_current_device())
           : stream(get_a_stream(group, dev)), device(dev)
       { }
 
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -115,6 +115,10 @@ namespace resources
           return count;
         }());
 
+        if (dev < 0) {
+          dev = get_current_device();
+        }
+
         std::call_once(devices[dev].onceFlag, [&] {
           auto d{device_guard(dev)};
           if (devices[dev].streams[0] == nullptr) {
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -18,6 +18,7 @@ http://github.com/llnl/camp
 #ifdef CAMP_ENABLE_CUDA
 
 #include <cuda_runtime.h>
+#include <mutex>
 
 namespace camp
 {
@@ -113,10 +114,9 @@ namespace resources
         });
 
         if (num < 0) {
-          devices[dev].mtx.lock();
+          std::lock_guard<std::mutex> guard(devices[dev].mtx);
           devices[dev].previous = (devices[dev].previous + 1) % num_streams;
           num = devices[dev].previous;
-          devices[dev].mtx.unlock();
         } else {
           num = num % num_streams;
         }
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -87,6 +87,17 @@ namespace resources
         return dev;
       }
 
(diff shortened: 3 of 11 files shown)


include/camp/resource/cuda.hpp
```diff
@@ -79,34 +79,53 @@ namespace resources
...
+              campcudaErrchk(cudaStreamCreate(&s));
             }
           }
         });
 
         if (num < 0) {
-          m_mtx.lock();
-          previous = (previous + 1) % 16;
-          m_mtx.unlock();
-          return streams[previous];
```
MrBurmark, line 102: This also fixes a race condition here.

davidbeckingsale (dismissed):

include/camp/resource/cuda.hpp
```diff
@@ -129,20 +148,18 @@ namespace resources
...
+      Cuda(int group = -1, int dev = get_current_device())
+          : stream(get_a_stream(group, dev)), device(dev)
       {
       }
 
       /// Create a resource from a custom stream
       /// The device specified must match the stream, if none is specified the
       /// currently selected device is used.
-      static Cuda CudaFromStream(cudaStream_t s, int dev = -1)
+      static Cuda CudaFromStream(cudaStream_t s, int dev = get_current_device())
```
trws, line 161: All the rest of this seems good to me. If we're going to support multiple devices this way though, this should be pulling the device out of the context associated with the stream so we don't end up with a mismatched pair.

MrBurmark: Good idea, I haven't used the driver API before, but this looks pretty straight forward to accomplish.

MrBurmark: That can be done with cuda, but I don't see how I can find the device of a hip stream.

trws: Take a look at https://rocm-developer-tools.github.io/HIP/classihipStream__t.html

It has a `getDevice` member function that returns an internal struct of theirs with a device id in it.

MrBurmark: It looks like rocm 5 has a hipGetStreamDeviceId function, I'll give that a try https://docs.amd.com/bundle/HIP-API-Guide-v5.3/page/a00196.html#gad16a23519ee4d7d87c7be53004b76608.

include/camp/resource/hip.hpp
```diff
@@ -75,34 +78,57 @@ namespace resources
...
+          num = num % num_streams;
         }
 
-        return streams[num % 16];
+        return devices[dev].streams[num];
       }
 
       // Private from-stream constructor
-      Hip(hipStream_t s, int dev = 0) : stream(s), device(dev) {}
+      Hip(hipStream_t s, int dev) : stream(s), device(dev) {}
```
trws, line 131: I see you wired in the way to get the device from the stream probably use that here as well?

trws (commented):
One comment on Hip, also it looks like we don't have a Sycl constructor that takes a sycl::queue, it would probably make sense to add that so it all matches, and they come with a `.get_device()` method as well.

MrBurmark:
@trws We ran into an issue in RAJA where we had a static storage duration resource whose constructor was called before main. With these changes it won't be possible to create a resource without it calling runtime functions. I propose adding something like this as a remedy.
```
static Cuda CudaStream0(int dev = get_current_device()) { return {0, dev}; }
```

include/camp/resource/hip.hpp
MrBurmark: That makes some sense, but take a look at what I say about constructing a resource without any runtime calls.

trws:
What happens on the runtime call, is it a crash or an error code? If it’s an error code, we could have it default to stream zero in that case. As for this, I think having this function with its dev argument is actually a reasonable solution, specifying any non-default device could mean “I already know, don’t check” as long as we swap the zero default for a sentinel value like -1 or something. What do you think?


---
Sent from Workspace ONE Boxer<https://whatisworkspaceone.com/boxer>

On February 21, 2023 at 4:21:54 PM PST, Jason Burmark ***@***.***> wrote:

@MrBurmark commented on this pull request.

________________________________

In include/camp/resource/hip.hpp<https://urldefense.us/v3/__https://github.com/LLNL/camp/pull/125*discussion_r1113700961__;Iw!!G2kpM7uM-TzIFchu!lYDYQLgkW9X-dczwnO61mXW74IcIGqICFogxw8uT6muwBqvADbWKMRbbxZNy1bvgRg$>:

>        }

       // Private from-stream constructor
-      Hip(hipStream_t s, int dev = 0) : stream(s), device(dev) {}
+      Hip(hipStream_t s, int dev) : stream(s), device(dev) {}


That makes some sense, but take a look at what I say about constructing a resource without any runtime calls.

MrBurmark:
The symptom is that later kernel calls fail with an error something like couldn't find kernel.
I'm happy using -1 as the default argument here and not checking if the argument is non-negative.

MrBurmark:
Random other thought, should the Cuda, Hip, and Sycl constructors be explicit?

include/camp/resource/hip.hpp
MrBurmark: I think you're saying here that it makes more sense to put the conditional and get_device_from_stream call here rather than in HipFromStream. I'll do that.

rhornung67:
@MrBurmark it's @trws and your call, but I think all single arg ctors should be explicit to avoid unwanted type conversions.

trws:
Oh man, yes.  I must not have been thinking about it but even the double argument ones have defaults and need it as well.

rhornung67 (approved):
I think this looks good. Do any tests need to be added in camp or RAJA?

MrBurmark:
@trws I'm a bit confused, on LC I have to manually link the resource tests because of the libcuda.so dependency, but the CI appears to be passing here.

MrBurmark:
Windows appears to be having a problem with throw_re that I don't understand.

trws:
Oddly it looks like it's causing a linkage issue on windows, I'm in a poor position to debug it but will at least take a look this morning.

MrBurmark:
Trying out something for the windows build, hopefully its something like what you were suggesting.

CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ endif()
 
 if (COMPILER_FAMILY_IS_MSVC)
   if (NOT BUILD_SHARED_LIBS)
-    target_compile_definitions(camp PUBLIC WIN_STATIC_BUILD)
+    target_compile_definitions(camp PUBLIC CAMP_WIN_STATIC_BUILD)
```
trws, line 134: That looks great, and it seems like it works, one tweak to clean up something right here and I think this one is good to go.

```suggestion
    target_compile_definitions(camp PRIVATE CAMP_WIN_STATIC_BUILD)
```

trws (changes requested):

trws:
And it looks like I broke it, that's somewhat disturbing.  If you go back to public, I'll take it and we can watch it downstream.  This shouldn't block getting the overall really good change in.

trws (dismissed):

include/camp/config.in.hpp
```diff
@@ -5,6 +5,8 @@
 #cmakedefine CAMP_ENABLE_CUDA
 #cmakedefine CAMP_ENABLE_HIP
 #cmakedefine CAMP_ENABLE_SYCL
+#cmakedefine CAMP_WIN_STATIC_BUILD
+#cmakedefine CAMP_DLL_EXPORTS
```
rhornung67, line 9: Is this going to be confusing with the 'CAMP_DLL_EXPORT' macro constant defined below?

include/camp/resource/cuda.hpp
```diff
@@ -129,20 +187,19 @@ namespace resources
...
-      /// currently selected device is used.
+    public:
+      explicit Cuda(int group = -1, int dev = get_current_device())
+          : stream(get_a_stream(group, dev)), device(dev)
+      { }
+
+      /// Create a resource from a custom stream.
+      /// If device is specified it must match the stream. If device is
+      /// unspecified, we will get it from the stream.
+      /// This may be called before main if device is specified as no calls to
```
rhornung67, line 199: We had an issue recently that took a while to resolve where we were making Cuda runtime library calls before entering main(). Will this present similar problems, potentially?

include/camp/resource/host.hpp
```diff
@@ -32,7 +35,7 @@ namespace resources
     class Host
     {
     public:
-      Host(int /* group */ = -1) {}
+      explicit Host(int /* group */ = -1) {}
```
rhornung67, line 38: Why does this take an arg if it's not used?

rhornung67 (commented):
A few questions.

include/camp/resource/host.hpp
trws: Consistency, without it the type erasing wrapper wouldn't compile.

include/camp/config.in.hpp
trws: The DLL_EXPORTS one shouldn't be exported here, we want that so that we export while compiling ourself and import while compiling others.

include/camp/resource/cuda.hpp
MrBurmark: This coding was the source of that problem. We ended up changing the way that we create some resources to avoid the issue that works before and after these changes.

include/camp/config.in.hpp
MrBurmark: Wait, should this cmakedefine be removed then?

rhornung67: Not sure. I will defer to you folks.

rhornung67 (dismissed):

include/camp/config.in.hpp
trws: I think so, the first one we need, the second may actually cause problems downstream.

MrBurmark:
```suggestion
```

trws (approved):

include/camp/resource/cuda.hpp
trws: @MrBurmark, I just happened back across this.  Would it have solved the issue to have the streams bound lazily?

PR #127, opened 2023-04-12
Updates to SYCL backend

abagusetty:
Mostly minor cleanup to comply with SYCL spec and add a few features (extensions from oneapi)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -9,6 +9,8 @@ set(camp_VERSION_MINOR ${PROJECT_VERSION_MINOR})
 set(camp_VERSION_PATCH ${PROJECT_VERSION_PATCH})
 
 include(CheckCXXCompilerFlag)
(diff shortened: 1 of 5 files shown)


trws:
This looks really good, thanks for the contribution @abagusetty!  Do you have some more extensions in mind before moving this forward?

abagusetty:
> This looks really good, thanks for the contribution @abagusetty! Do you have some more extensions in mind before moving this forward?

Hi @trws Thanks for taking a peek. I just wanted to make sure this works on (a) Intel PVC (a few combinations) (b) get in touch with @homerdin related to a use case (c) a bit extended -- run on CUDA and HIP devices (d) Do a sanity check with Umpire. Shouldn't be more than day or so. Will gently ping you, when ready. Thanks again!

trws:
Sounds great, thanks!

PR #133, opened 2023-06-01
Make get_platform a const method

MrBurmark:
There may not be many things you can do with a const resource but getting the platform seems reasonable.


trws (approved):
Sounds right to me.  Here's looking forward to C++20 when we can mark that one constexpr too. 😁

PR #135, opened 2023-07-19
Add missing header to cuda resource

adayton1:
* Several resources were missing the <mutex> include
* Some other headers do not need to be included if certain configurations are not enabled


trws (dismissed):
Good stuff, thanks @adayton1!

trws:
Bah, looks like the ubuntu bionic-based containers are no longer supported, I'll pop up a commit here so we can get the build tests run.

MrBurmark (approved):
We should probably do the same for the other resources.

adayton1:
> We should probably do the same for the other resources.

Good call. Done.

MrBurmark (approved):

trws:
@adayton1, would you mind doing a merge or rebase here with the updated test containers so we can get the CI cuda builds going on this? I'm happy with it but would prefer to see those green before a merge.

trws:
This branch is configured not to allow maintainers to merge or rebase, would be happy to move this forward if we can get the tests going.

adayton1:
Sorry, I’ve been on vacation. I just merged in main, so hopefully that will take care of it. I’ll be back on vacation this afternoon until next week.

trws:
All passed, thanks @adayton1!

PR #137, opened 2023-07-19
Don't zero in Resource::allocate

MrBurmark:
I ran into an issue where I setting pinned memory on the host immediately allocating it failed intermittently.
I discovered that Resource::allcoate actually calls the underlying calloc instead of allocate.


adayton1 (approved):
LGTM

trws (dismissed):

trws (approved):

PR #140, opened 2023-10-16
Allow for structured bindings from const camp::tuple

### LINKED ISSUE #139: implementation of `tuple_element<i,camp::tuple<T...>>` not allowing structured binding using `const auto`
Hello,
In the StructuredBindings unit test, if I change the return type to `const auto` as such:
```c++
TEST(CampTuple, StructuredBindings)
{
  auto t = camp::make_tuple(3, 9.9);
  const auto [a, b] = t;
  ASSERT_EQ(a, 3);
  ASSERT_NEAR(b, 9.9, 1e-15);
}
```
I get the following error:
```
[ 34%] Building CXX object test/CMakeFiles/tuple.dir/tuple.cpp.o
cd /usr/workspace/settgast/Codes/geosx/camp/build/test && /usr/lib64/ccache/c++ -DGTEST_HAS_DEATH_TEST=0 -I/usr/workspace/settgast/Codes/geosx/camp/include -I/usr/workspace/settgast/Codes/geosx/camp/build/include -isystem /usr/workspace/settgast/Codes/geosx/camp/extern/blt/thirdparty_builtin/googletest/googletest/include -Wall -Wextra      -fPIE -std=c++17 -MD -MT test/CMakeFiles/tuple.dir/tuple.cpp.o -MF CMakeFiles/tuple.dir/tuple.cpp.o.d -o CMakeFiles/tuple.dir/tuple.cpp.o -c /usr/workspace/settgast/Codes/geosx/camp/test/tuple.cpp
/usr/workspace/settgast/Codes/geosx/camp/test/tuple.cpp: In member function ‘virtual void CampTuple_StructuredBindings_Test::TestBody()’:
/usr/workspace/settgast/Codes/geosx/camp/test/tuple.cpp:362:23: error: binding reference of type ‘int&&’ to ‘camp::tuple_element_t<0, camp::tuple<int, double> >’ {aka ‘const int’} discards qualifiers
   const auto [a, b] = t;
                       ^
/usr/workspace/settgast/Codes/geosx/camp/test/tuple.cpp:362:23: error: binding reference of type ‘double&&’ to ‘camp::tuple_element_t<1, camp::tuple<int, double> >’ {aka ‘const double’} discards qualifiers
```

If I change the `camp::tuple` to `std::tuple`, it compiles fine.

Also...in tuple.hpp if I replace the implementation:
```c++
  template <size_t i, typename ... T>
  struct tuple_element<i, camp::tuple<T...>> {
    using type = decltype(camp::get<i>(camp::tuple<T...>{}));
  };
```
with something I took/altered from [cppreference](https://en.cppreference.com/w/cpp/utility/tuple/tuple_element):
```c++
  // recursive case
  template<std::size_t I, class Head, class... Tail>
  struct tuple_element<I, camp::tuple<Head, Tail...>>
      : std::tuple_element<I - 1, camp::tuple<Tail...>>
  { };
 
// base case
  template<class Head, class... Tail>
  struct tuple_element<0, camp::tuple<Head, Tail...>>
  {
      using type = Head;
  };
```
it compiles fine.

@trws @adayton1 Is this an OK change to make? I can submit a PR with the changes and add a test for the const case if so. Also should probably add testing for all the structured binding cases: `const auto`, `const auto &`, `const auto &&`?

rrsettgast:
closes #139

### DIFF
--- a/test/tuple.cpp
+++ b/test/tuple.cpp
@@ -360,8 +360,8 @@ TEST(CampTuple, StructuredBindings)
 {
   auto t = camp::make_tuple(3, 9.9);
   auto [a, b] = t;
-  static_assert(std::is_same_v< decltype(a), int > );
-  static_assert(std::is_same_v< decltype(b), double > );
+  static_assert(std::is_same< decltype(a), int >::value );
+  static_assert(std::is_same< decltype(b), double >::value );
   ASSERT_EQ(a, 3);
   ASSERT_NEAR(b, 9.9, 1e-15);
 }
--- a/test/tuple.cpp
+++ b/test/tuple.cpp
@@ -360,9 +360,53 @@ TEST(CampTuple, StructuredBindings)
 {
   auto t = camp::make_tuple(3, 9.9);
   auto [a, b] = t;
+  static_assert(std::is_same_v< decltype(a), int > );
+  static_assert(std::is_same_v< decltype(b), double > );
   ASSERT_EQ(a, 3);
   ASSERT_NEAR(b, 9.9, 1e-15);
 }
+
+TEST(CampTuple, StructuredBindingsConst)
+{
+  auto t = camp::make_tuple(3, 9.9);
+  const auto [a, b] = t;
+  static_assert(std::is_same_v< decltype(a), const int > );
+  static_assert(std::is_same_v< decltype(b), const double > );
+  ASSERT_EQ(a, 3);
+  ASSERT_NEAR(b, 9.9, 1e-15);
+}
+
+TEST(CampTuple, StructuredBindingsRef)
+{
+  auto t = camp::make_tuple(3, 9.9);
+  auto & [a, b] = t;
+  static_assert(std::is_same< decltype(a), int >::value ); // Not an int &
+  static_assert(std::is_same< decltype(b), double >::value ); // Not a double &
+  std::cout<<typeid(a).name()<<std::endl;
+  ASSERT_EQ(a, 3);
+  ASSERT_NEAR(b, 9.9, 1e-15);
+
+  a = 4; // Well...kind of an int &
+  b = 10.1; // also kind of a double &
+  ASSERT_EQ(camp::get<0>(t), 4);
+  ASSERT_NEAR(camp::get<1>(t), 10.1, 1e-15);
+}
+
+TEST(CampTuple, StructuredBindingsRefConst)
+{
+  auto t = camp::make_tuple(3, 9.9);
+  const auto & [a, b] = t;
+  static_assert(std::is_same< decltype(a), const int >::value );  // Not an const int &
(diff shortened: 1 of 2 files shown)


test/tuple.cpp
```diff
@@ -360,9 +360,53 @@ TEST(CampTuple, StructuredBindings)
 {
   auto t = camp::make_tuple(3, 9.9);
   auto [a, b] = t;
+  static_assert(std::is_same_v< decltype(a), int > );
+  static_assert(std::is_same_v< decltype(b), double > );
```
trws, line 364: Unfortunately, we currently still have to support c++14, so these need to downgrade. Otherwise this all looks great.

trws (changes requested):

test/tuple.cpp
```diff
@@ -360,9 +360,53 @@ TEST(CampTuple, StructuredBindings)
...
+  const auto & [a, b] = t;
+  static_assert(std::is_same< decltype(a), const int >::value );  // Not an const int &
+  static_assert(std::is_same< decltype(b), const double >::value ); // Not a const double &
+  std::cout<<typeid(a).name()<<std::endl;
+  ASSERT_EQ(a, 3);
+  ASSERT_NEAR(b, 9.9, 1e-15);
+
+  // these should fail to compile.
+  // a = 4;
+  // b = 10.1;
```
rrsettgast, line 407: @trws Do you have a preferred method for testing that compilations fail, or are you ok not testing? I don't think you can protect yourself fully given the `tuple` of `references` behavior?

rrsettgast:
@trws I assume you squash merge, so this history doesn't need to be clean?

test/tuple.cpp
trws: I have some infrastructure for compilation death tests, but pulled it out of mainline camp because checking the error messages proved rather brittle. If possible I'd say factor the test to use decltype on the binding to verify it's const-qualified as it should be, then we can keep the tests positive. If not I can dredge that stuff back up. It amounts to generating a non-auto cmake target that builds the test then registering cmake --build --target tgt as the test for ctest to run. Surprisingly robust, but also frustrating to make sure you're testing only what you want to test.

trws:
Yeah that’s no problem.

test/tuple.cpp
rrsettgast: Well...maybe the `static_assert( is_same<>)` checks are sufficient. It is up to you.

trws: I think the static_asserts are sufficient.

rrsettgast:
@trws Is there something I need to do to pass the checks that are failing?

trws:
Nope, not this time.  The sycl container broke a couple of days ago and I haven't had time to fix it yet. Thanks @rrsettgast!

trws (approved):

PR #141, opened 2023-10-21
Avoid some -Wshadow warnings in gcc12

rrsettgast:
When I was compiling camp with gcc12, I get the several of the following types of errors:
```
In file included from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/camp.hpp:24,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/src/common/types.hpp:12,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/src/discretizations/unitTests/testLagrangeBasis.cpp:6:
/usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/tuple.hpp: In function ‘constexpr camp::tuple_element_t<index, camp::tuple<Types ...> >& camp::get(const tuple<Types ...>&)’:
/usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/tuple.hpp:440:29: error: declaration of ‘t’ shadows a global declaration [-Werror=shadow]
  440 | get(const tuple<Types...>&  t) noexcept
      |     ~~~~~~~~~~~~~~~~~~~~~~~~^
In file included from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/value.hpp:14,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/number/if.hpp:14,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/number.hpp:15,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/list/list.hpp:14,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/list/at.hpp:16,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/lambda.hpp:17,
                 from /usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/camp.hpp:18:
/usr/workspace/settgast/Codes/geosx/Shiva/tpl/camp/include/camp/number/number.hpp:40:7: note: shadowed declaration is here
   40 | using t = num<true>;
      |       ^
```
I think that while this shouldn't be an actual problem, it is a valid warning. Changing the name of the parameter from `t` to `tt` avoids the shadowing.

edit: I am aware that I should be using `target_include_directories` with the `SYSTEM` specifier. I had made a mistake and didn't include `SYSTEM`, but I still think that this could be addressed in `camp`.

edit2: I figured out the build system is trying to compile the error.cpp file in camp. This is where the warning is coming up. I added the `-Wno-shadow` flag to the `camp` target, and this removes the warning. So there are workarounds.

### DIFF
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -437,143 +437,143 @@ struct as_list_s<tagged_tuple<camp::list<Tags...>, Args...>> {
 // by index
 template <camp::idx_t index, class... Types>
 CAMP_HOST_DEVICE constexpr tuple_element_t<index, tuple<Types...>> const&
-get(const tuple<Types...>&  t) noexcept
+get(const tuple<Types...>&  tt) noexcept
 {
   using internal::tpl_get_store;
   static_assert(tuple_size<tuple<Types...>>::value > index, "index out of range");
-  return static_cast<const tpl_get_store<tuple<Types...>, index>&>(t.base).get_inner();
+  return static_cast<const tpl_get_store<tuple<Types...>, index>&>(tt.base).get_inner();
 }
 
 template <camp::idx_t index, class... Types>
 CAMP_HOST_DEVICE constexpr tuple_element_t<index, tuple<Types...>> const&&
-get(const tuple<Types...>&& t) noexcept
+get(const tuple<Types...>&& tt) noexcept
 {
   using internal::tpl_get_store;
   static_assert(tuple_size<tuple<Types...>>::value > index, "index out of range");
-  return static_cast<const tpl_get_store<tuple<Types...>, index>&&>(t.base).get_inner();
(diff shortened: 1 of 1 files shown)


trws:
I'd be good with working around this, really should probably be using prefixed names or something even but it's worth doing.  It looks like all the checks were cancelled, and I'm not sure why.  Is this ready for testing?

rrsettgast:
> I'd be good with working around this, really should probably be using prefixed names or something even but it's worth doing. It looks like all the checks were cancelled, and I'm not sure why. Is this ready for testing?

I don't know why they were cancelled? I think it is ready. It is a trivial change. If you prefer prefixing the `using t=num<true>` then I can do that instead of what I did here.

trws (approved):

PR #143, opened 2023-10-26
use hipPointerAttribute_t.type as HIP is removing hipPointerAttribute_t.memoryType

rakesroy:
- In ROCm6.0 hipPointerAttribute_t.memoryType will be removed from HIP.
- Instead hipPointerAttribute_t.type to be used.
- This is causing build failure for RAJAPerf

### DIFF
--- a/include/camp/resource/hip.hpp
+++ b/include/camp/resource/hip.hpp
@@ -113,7 +113,11 @@ namespace resources
         hipPointerAttribute_t a;
         hipError_t status = hipPointerGetAttributes(&a, p);
         if (status == hipSuccess) {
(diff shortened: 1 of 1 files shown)


trws:
Thanks for the PR @rakesroy.  I'm a bit concerned that `.type` isn't documented for ROCM versions below 6, do we need an ifdef here for pre-6 versions or is it just undocumented in 5.*?  The changelog looks like it's changing, so I'm guessing we need the ifdef or similar to maintain compatibility across versions right?

rakesroy:
@trws 
I have updated this PR to to use .type when (HIP_VERSION_MAJOR >= 6), otherwise to use .memoryType for backwards compatibility.

PR #146, opened 2024-03-05
Add std::array-like class

adayton1:
* Add camp::array as a portable alternative to std::array
* Add portable testing infrastructure
* Register gtests with ctest for filtering (which requires CMake 3.10)


artv3:
I was recently asked if we will now expect users to call camp directly in their applications. Will we? or do we think we will add camp features under a raja namespace?

trws:
The traditional thing has been to wrap everything in raja namespace, but I think the big decision will be whether we do it for desul or not, and for some things here.  We could make it so that camp things can be re-exported or only exported into a different namespace, but I'm not sure which is better to be honest.

trws:
This looks pretty slick @adayton1, let me know when you're ready for a review and discussion.

adayton1:
@trws this

> This looks pretty slick @adayton1, let me know when you're ready for a review and discussion.

It's ready for review!

test/array.cpp
```diff
@@ -0,0 +1,400 @@
...
+          !(a != a);
+} CAMP_TEST_END(array, not_equal)
+
+CAMP_TEST_BEGIN(array, less_than)
+{
+   camp::array<int, 2> a{1, 2};
+   camp::array<int, 2> b{1, 3};
+
+   return a < b &&
+          !(b < a);
```
MrBurmark, line 282: Should test the case when they are equals as well.
```suggestion
   return !(a < a) &&
          a < b &&
          !(b < a);
```

test/array.cpp
```diff
@@ -0,0 +1,400 @@
...
+          a <= b &&
+          !(b <= a);
+} CAMP_TEST_END(array, less_than_or_equal)
+
+CAMP_TEST_BEGIN(array, greater_than)
+{
+   camp::array<int, 2> a{1, 2};
+   camp::array<int, 2> b{1, 3};
+
+   return b > a &&
```
MrBurmark, line 300: equal case

adayton1: Good idea. Done.

include/camp/array.hpp
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      return false;
+   }
+
+   template <class T, std::size_t N>
+   CAMP_HOST_DEVICE constexpr bool operator<=(const array<T, N>& lhs,
+                                              const array<T, N>& rhs) {
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] < rhs[i]) {
```
rhornung67, line 172:
```suggestion
         if (lhs[i] <= rhs[i]) {
```

include/camp/array.hpp
```diff
@@ -0,0 +1,291 @@
...
+   }
+
+   template <class T, std::size_t N>
+   CAMP_HOST_DEVICE constexpr bool operator<(const array<T, N>& lhs,
+                                             const array<T, N>& rhs) {
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] < rhs[i]) {
+            return true;
+         }
+         else if (lhs[i] == rhs[i]) {
```
rhornung67, line 157: If the `==` check is true, doesn't that mean that the `<` operator should return false?

include/camp/array.hpp
```diff
@@ -0,0 +1,291 @@
...
+   }
+
+   template <class T, std::size_t N>
+   CAMP_HOST_DEVICE constexpr bool operator>(const array<T, N>& lhs,
+                                             const array<T, N>& rhs) {
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] > rhs[i]) {
+            return true;
+         }
+         else if (lhs[i] == rhs[i]) {
```
rhornung67, line 193: Same question as above regarding the `==` check.

include/camp/array.hpp
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      return false;
+   }
+
+   template <class T, std::size_t N>
+   CAMP_HOST_DEVICE constexpr bool operator>=(const array<T, N>& lhs,
+                                              const array<T, N>& rhs) {
+      for (std::size_t i = 0; i < N; ++i) {
+         if (lhs[i] > rhs[i]) {
```
rhornung67, line 208:
```suggestion
         if (lhs[i] >= rhs[i]) {
```

Similar question as in the `<=` operator case.

include/camp/array.hpp
```diff
@@ -0,0 +1,291 @@
...
+   CAMP_HOST_DEVICE constexpr const T& get(const array<T, N>& a) noexcept {
+      return a[I];
+   }
+
+   template <std::size_t I, class T, std::size_t N>
+   CAMP_HOST_DEVICE constexpr const T&& get(const array<T, N>&& a) noexcept {
+      return move(a[I]);
+   }
+
+   // TODO: Investigate the implementation from cppreference
```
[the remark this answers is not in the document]
adayton1: I'm having a little trouble with the implementation:

```
namespace detail {
      template <class T, std::size_t N, std::size_t... I>
      CAMP_HOST_DEVICE constexpr array<std::remove_cv_t<T>, N>
         to_array_impl(T (&a)[N], idx_seq<I...>)
      {
         return {{a[I]...}};
      }

      template <class T, std::size_t N, std::size_t... I>
      CAMP_HOST_DEVICE constexpr array<std::remove_cv_t<T>, N>
         to_array_impl(T (&&a)[N], idx_seq<I...>)
      {
         return {{move(a[I])...}};
      }
   }

   template <class T, std::size_t N>
   CAMP_HOST_DEVICE constexpr array<std::remove_cv_t<T>, N>
      to_array(T (&a)[N])
   {
      return detail::to_array_impl(a, make_idx_seq_t<N>{});
   }

   template <class T, std::size_t N>
   CAMP_HOST_DEVICE constexpr array<std::remove_cv_t<T>, N>
      to_array(T (&&a)[N])
   {
      return detail::to_array_impl(move(a), make_idx_seq_t<N>{});
   }
```

Test case:

```
int temp[3] = {1, 2, 10};
camp::array<int, 3> a = camp::to_array(temp);
camp::array<int, 3> b = camp::to_array(camp::move(temp));
```

Error messages (there seem to be issues with non-type template parameters):

```
In file included from camp/test/array.cpp:1:
camp/include/camp/array.hpp:263:14: error: no matching function for call to 'to_array_impl'
      return detail::to_array_impl(a, make_idx_seq_t<N>{});
             ^~~~~~~~~~~~~~~~~~~~~
camp/test/array.cpp:351:34: note: in instantiation of function template specialization 'camp::to_array<int, 3UL>' requested here
   camp::array<int, 3> a = camp::to_array(temp);
                                 ^
camp/include/camp/array.hpp:246:10: note: candidate template ignored: substitution failure [with T = int, N = 3]: deduced non-type template argument does not have the same type as the corresponding template parameter ('long' vs 'std::size_t' (aka 'unsigned long'))
         to_array_impl(T (&a)[N], idx_seq<I...>)
         ^
camp/include/camp/array.hpp:253:10: note: candidate template ignored: substitution failure [with T = int, N = 3]: deduced non-type template argument does not have the same type as the corresponding template parameter ('long' vs 'std::size_t' (aka 'unsigned long'))
         to_array_impl(T (&&a)[N], idx_seq<I...>)
         ^
camp/include/camp/array.hpp:270:14: error: no matching function for call to 'to_array_impl'
      return detail::to_array_impl(move(a), make_idx_seq_t<N>{});
             ^~~~~~~~~~~~~~~~~~~~~
camp/test/array.cpp:352:34: note: in instantiation of function template specialization 'camp::to_array<int, 3UL>' requested here
   camp::array<int, 3> b = camp::to_array(camp::move(temp));
                                 ^
camp/include/camp/array.hpp:246:10: note: candidate template ignored: substitution failure [with T = int, N = 3]: deduced non-type template argument does not have the same type as the corresponding template parameter ('long' vs 'std::size_t' (aka 'unsigned long'))
         to_array_impl(T (&a)[N], idx_seq<I...>)
         ^
camp/include/camp/array.hpp:253:10: note: candidate template ignored: substitution failure [with T = int, N = 3]: deduced non-type template argument does not have the same type as the corresponding template parameter ('long' vs 'std::size_t' (aka 'unsigned long'))
         to_array_impl(T (&&a)[N], idx_seq<I...>)
         ^
2 errors generated.
```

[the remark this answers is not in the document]
adayton1: If I use `auto N` and `auto... I` instead of `std::size_t N` and `std::size_t... I`, it compiles fine. But I think the use of `auto` for non-type template parameters isn't supported until C++17.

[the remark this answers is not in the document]
adayton1: I can conditionally use the implementation with auto based on whether C++17 is being used.

include/camp/array.hpp
```diff
@@ -0,0 +1,324 @@
...
+         }
+         else if (lhs[i] == rhs[i]) {
+            continue;
+         }
+         else {
+            return false;
+         }
+      }
+
+      return false;
```
long58, line 166:
```suggestion
      for (std::size_t i = 0; i < N; ++i) {
         if (lhs[i] >= rhs[i]) {
            return false;
         }
      }

      return true;
```
Using operator== as an example, shouldn't this operator looks more like this? 
And same thing for the other operators <=, >, >=

include/camp/array.hpp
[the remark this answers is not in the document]
MrBurmark: I think camp has its own idx_t or something like that that should be used with make_idx_seq_t.

include/camp/array.hpp
adayton1: If it's equal, then we still need to compare the next element in the array.

include/camp/array.hpp
adayton1: Maybe there is a specific ordering that I'm missing, but I was thinking if you had two arrays, a = {1, 2} and b = {1, 3}, then a < b even though the first element is equal.

include/camp/array.hpp
adayton1: I think it depends on the ordering. I might be missing something. Should {1, 3} > {1, 2} return true or false?

rhornung67: It returns true. See the code example I posted in a separate comment.

adayton1: I've modified the implementation to match std::lexicographical_compare: https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare

long58: hmmm yeah, I guess that should return true... I wasn't aware (until now) that it was a lexicographical compare and not just a raw element-wise comparison.

rhornung67: @adayton1 I think you have everything correct. I should have verified before posting my comments. Sorry for the churn.

adayton1:
@rhornung67 @long58 I've modified the comparison operators. operator< now matches the implementation in std::lexicographical_compare (https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare) and all the other operators are implemented in terms of operator< or operator=.

Alternatively, I'm not really sure if there's a benefit to having comparison operators besides == and !=. I could potentially remove the others.

include/camp/array.hpp
```diff
@@ -0,0 +1,287 @@
...
+      using size_type = std::size_t;
+      using difference_type = std::ptrdiff_t;
+      using reference = value_type&;
+      using const_reference = const value_type&;
+      using pointer = value_type*;
+      using const_pointer = const value_type*;
+      using iterator = pointer;
+      using const_iterator = const_pointer;
+
+      // TODO: Investigate device trap
```
rhornung67, line 35: @adayton1 FWIW, we use this macro in RAJA to handle the various back-end cases https://github.com/LLNL/RAJA/blob/develop/include/RAJA/util/macros.hpp#L143

rhornung67:
> @rhornung67 @long58 I've modified the comparison operators. operator< now matches the implementation in std::lexicographical_compare (https://en.cppreference.com/w/cpp/algorithm/lexicographical_compare) and all the other operators are implemented in terms of operator< or operator=.
> 
> Alternatively, I'm not really sure if there's a benefit to having comparison operators besides == and !=. I could potentially remove the others.

For consistency with std::array, I think you should keep them.

include/camp/array.hpp
rhornung67: @long58 I also learned the same lesson. 😄

include/camp/array.hpp
adayton1: I will look into that, but I don't want that to hold up this pull request.

include/camp/array.hpp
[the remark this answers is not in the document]
adayton1: But the std::array class prescribes std::size_t as the index type, so I'm not sure I want to go that route.

rhornung67 (dismissed):
Thank you @adayton1

long58 (approved):

adayton1:
The Docker rocm tests are failing, I think probably because the machine doesn't have any AMD GPUs. Should I turn off the testing and just do the build? Or maybe I could try to build in a filter so only the host tests run.

I also noticed that the docker nvcc builds don't actually build with nvcc - they are host only builds.

include/camp/array.hpp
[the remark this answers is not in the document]
MrBurmark: Try this https://github.com/LLNL/camp/blob/79c320fa09db987923b56884afdc9f82f4b70fc4/include/camp/number.hpp#L117

rhornung67:
> The Docker rocm tests are failing, I think probably because the machine doesn't have any AMD GPUs. Should I turn off the testing and just do the build? Or maybe I could try to build in a filter so only the host tests run.
> 
> I also noticed that the docker nvcc builds don't actually build with nvcc - they are host only builds.

I believe all of our projects that run CI on cloud services only build and do not run anything for GPU jobs.  Was it running before your PR? If so, I don't know how that would work.

adayton1:
> > The Docker rocm tests are failing, I think probably because the machine doesn't have any AMD GPUs. Should I turn off the testing and just do the build? Or maybe I could try to build in a filter so only the host tests run.
> > I also noticed that the docker nvcc builds don't actually build with nvcc - they are host only builds.
> 
> I believe all of our projects that run CI on cloud services only build and do not run anything for GPU jobs. Was it running before your PR? If so, I don't know how that would work.

Yeah, they do run tests. I'm not sure how that was working before, either. The nvcc builds were actually host only, so tests would have run fine there. But the rocm one would be broken. For now, I've filtered out all the tests that run code on GPUs. We'll want to clean up this CI in a follow on merge request.

include/camp/array.hpp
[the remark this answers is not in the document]
adayton1: Thanks, Jason. That worked.

rhornung67 (approved):

PR #148, opened 2024-04-12
Add a way to compare resources

kab163:
This compare function will compare 2 different resources. 

I need this because Umpire would like to add a resource aware pool, but in order to do that we need to know if the resources that are being passed into the pool per allocate call are the same or different. (We handle each case differently.) 

- [x] Double check that my Omp compare is sufficient (see TODO statement)
- [x] Add Doxygen comments to compare functions
- [x] Test this with my resource aware pool prototype - this will act as another check to see if the compare function behave as expected.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -127,7 +127,7 @@ namespace resources
       bool operator==(T const& t) const
       {
         if(get_platform() == t.get_platform()) { 
-          return get<T>().compare(t);
+          return get<T>() == t;
         }
         return false;
       }
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -123,22 +123,6 @@ namespace resources
         return !(*this == r);
       }
 
-      /*
-       * \brief Compares a generic Resource and a concrete Resource to see if they are equal. To find out, we
-       * first check to see if they are on the same platform. Then we call the operator == to compare the 
-       * type of the generic Resource (T) and the concrete one.
-       *
-       * \return True or false depending on comparison with m_value and t, assuming they are on the same platform.
-       */
-      template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
-      bool operator==(T const& t) const
-      {
-        if(get_platform() == t.get_platform()) { 
-          return get<T>() == t;
-        }
-        return false;
-      }
-
     private:
       class ContextInterface
       {
@@ -197,6 +181,22 @@ namespace resources
 
       std::shared_ptr<ContextInterface> m_value;
     };
+
+    /*
+     * \brief Compares a generic Resource and a concrete Resource to see if they are equal. To find out, we
+     * first check to see if they are on the same platform. Then we call the operator == to compare the 
+     * type of the generic Resource (T) and the concrete one.
+     *
+     * \return True or false depending on comparison with m_value and t, assuming they are on the same platform.
+     */
+    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
+    bool operator==(Resource const& r, T const& t)
+    {
+      if(r.get_platform() == t.get_platform()) { 
+        return r.get<T>() == t;
+      }
+      return false;
+    }
 
     template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
     bool operator==(T const& t, Resource const& r)
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -123,7 +123,7 @@ namespace resources
         return !(*this == r);
       }
 
-      template <typename T>
+      template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
       bool operator==(T const& t) const
       {
         if(get_platform() == t.get_platform()) { 
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -151,10 +151,7 @@ namespace resources
         ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
         Platform get_platform() const override { return m_modelVal.get_platform(); }
 
-        bool compare(Resource const& r) const override { 
-          T val = r.get<T>();
-          return m_modelVal == val; 
-        }
+        bool compare(Resource const& r) const override { return m_modelVal == r.get<T>(); }
 
         void *allocate(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.template allocate<char>(size, ma); }
         void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.calloc(size, ma); }
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -94,6 +94,12 @@ namespace resources
       void wait_for(Event *e) { m_value->wait_for(e); }
       void wait() { m_value->wait(); }
 
+      template <typename T>
+      bool compare (const camp::resources::Resource& r) {
+        auto result = dynamic_cast<ContextModel<T> *>(m_value.get());
+        return *result->compare(r);
+      }
+
     private:
       class ContextInterface
       {
@@ -225,6 +231,14 @@ namespace resources
 
       Res resource_;
     };
+
+      bool operator== (const camp::resources::Resource& l, const camp::resources::Resource& r)
+      {
+        if(l.get_platform() == r.get_platform()) {
+          return l.compare(r);
+        }
+        return false;
+      }
 
   }  // namespace v1
 }  // namespace resources
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -95,9 +95,10 @@ namespace resources
       void wait() { m_value->wait(); }
 
       template <typename T>
-      bool compare (const camp::resources::Resource& r) {
-        auto result = dynamic_cast<ContextModel<T> *>(m_value.get());
-        return *result->compare(r);
+      bool compare (camp::resources::Resource& r) {
+        auto this_result = dynamic_cast<T>(m_value.get());
+        auto r_result = dynamic_cast<T>(r.get());
+        return this_result.compare(r_result);
       }
 
     private:
@@ -232,13 +233,13 @@ namespace resources
       Res resource_;
     };
 
-      bool operator== (const camp::resources::Resource& l, const camp::resources::Resource& r)
-      {
-        if(l.get_platform() == r.get_platform()) {
-          return l.compare(r);
-        }
-        return false;
+    bool operator== (camp::resources::Resource& l, camp::resources::Resource& r)
+    {
+      if(l.get_platform() == r.get_platform()) {
+        return l.compare(r);
       }
+      return false;
+    }
 
   }  // namespace v1
 }  // namespace resources
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -95,7 +95,7 @@ namespace resources
       void wait() { m_value->wait(); }
 
       template <typename T>
-      bool compare (camp::resources::Resource& r) {
+      bool compare (const camp::resources::Resource& r) {
         auto this_result = dynamic_cast<T>(m_value.get());
         auto r_result = dynamic_cast<T>(r.get());
         return this_result.compare(r_result);
@@ -233,7 +233,7 @@ namespace resources
       Res resource_;
     };
 
-    bool operator== (camp::resources::Resource& l, camp::resources::Resource& r)
+    bool operator== (const camp::resources::Resource& l, const camp::resources::Resource& r)
     {
       if(l.get_platform() == r.get_platform()) {
         return l.compare(r);
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -188,37 +188,6 @@ namespace resources
       std::shared_ptr<ContextInterface> m_value;
     };
 
-    /*
-     * \brief Compares a generic Resource and a concrete Resource to see if they are equal. To find out, we
-     * first check to see if they are on the same platform. Then we call the operator == to compare the 
-     * type of the generic Resource (T) and the concrete one.
-     *
-     * \return True or false depending on comparison with m_value and t, assuming they are on the same platform.
-     */
-    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
-    bool operator==(Resource const& r, T const& t)
-    {
-      if(r.get_platform() == t.get_platform()) { 
-        return r.get<T>() == t;
-      }
-      return false;
-    }
-
-    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
-    bool operator==(T const& t, Resource const& r)
-    {
-      if(t.get_platform() == r.get_platform()) { 
-        return t == r.get<T>();
-      }
-      return false;
-    }
-
-    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
-    bool operator!=(T const& t, Resource const& r)
-    {
-      return !(t == r);
-    }
-
     template <Platform p>
     struct resource_from_platform;
     template <>
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -106,7 +106,7 @@ namespace resources
       bool operator==(Resource const& r)
       {
         if(get_platform() == r.get_platform()) {
-          return (*m_value == r);
+          return (m_value->compare(r));
         }
         return false;
       }
@@ -130,8 +130,7 @@ namespace resources
         virtual ~ContextInterface() {}
         virtual Platform get_platform() const = 0;
 
-        virtual bool operator==(Resource const& r) = 0;
-        virtual bool operator!=(Resource const& r) = 0;
+        virtual bool compare(Resource const& r) = 0;
 
         virtual void *allocate(size_t size, MemoryAccess ma = MemoryAccess::Device) = 0;
         virtual void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device) = 0;
@@ -152,8 +151,7 @@ namespace resources
         ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
         Platform get_platform() const override { return m_modelVal.get_platform(); }
 
-        bool operator==(Resource const& r) override { return m_modelVal == r.get<T>(); }
-        bool operator!=(Resource const& r) override { return m_modelVal != r.get<T>(); }
+        bool compare(Resource const& r) override { return m_modelVal == r.get<T>(); }
 
         void *allocate(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.template allocate<char>(size, ma); }
         void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.calloc(size, ma); }
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -198,6 +198,21 @@ namespace resources
       std::shared_ptr<ContextInterface> m_value;
     };
 
+    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
+    bool operator==(T const& t, Resource const& r)
+    {
+      if(t.get_platform() == r.get_platform()) { 
+        return t == r.get<T>();
+      }
+      return false;
+    }
+
+    template <typename T, std::enable_if_t<!std::is_same<T, Resource>::value>* = nullptr>
+    bool operator!=(T const& t, Resource const& r)
+    {
+      return !(t == r);
+    }
+
     template <Platform p>
     struct resource_from_platform;
     template <>
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -151,7 +151,10 @@ namespace resources
         ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
         Platform get_platform() const override { return m_modelVal.get_platform(); }
 
-        bool compare(Resource const& r) const override { return m_modelVal == r.get<T>(); }
+        bool compare(Resource const& r) const override { 
+          T val = r.get<T>();
+          return m_modelVal == val; 
+        }
 
         void *allocate(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.template allocate<char>(size, ma); }
         void *calloc(size_t size, MemoryAccess ma = MemoryAccess::Device) override { return m_modelVal.calloc(size, ma); }
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -94,6 +94,11 @@ namespace resources
       void wait_for(Event *e) { m_value->wait_for(e); }
       void wait() { m_value->wait(); }
 
+      /*
+       * \brief Compares two Resources to see if they are equal.
+       *
+       * \return True or false depending on comparison with m_value if they are on the same platform.
+       */
       bool operator==(Resource const& r)
       {
         if(get_platform() == r.get_platform()) {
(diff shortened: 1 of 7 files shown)


include/camp/resource/omp_target.hpp
```diff
@@ -192,6 +192,11 @@ namespace resources
         }
         return ret;
       }
+      bool compare (Omp* o)
+      {
+        return (dev == o->dev);
+        //TODO: do i need to compare addr???
```
trws, line 198: Yup, please do.  The value of the address is used by the OpenMP runtime to select a stream or equivalent.

include/camp/resource/cuda.hpp
```diff
@@ -259,6 +260,12 @@ namespace resources
       cudaStream_t get_stream() { return stream; }
       int get_device() { return device; }
 
+      bool compare (Cuda* c)
```
trws, line 263: Mostly this PR looks good, and like a nice addition, but the interface is slightly skewed between the type-erased and the strongly-typed classes. Would it be possible to either make the base class takes a pointer or, my preference, make each of these takes a reference so they match?  Having to sometimes do `.compare(&res)` and sometimes not seems unfortunate.

include/camp/resource.hpp
```diff
@@ -226,6 +239,11 @@ namespace resources
       Res resource_;
     };
 
+    bool operator== (camp::resources::Resource& l, camp::resources::Resource& r)
```
trws, line 242: Maybe `==` for strongly typed as well, so a strongly typed instance doesn't have to be wrapped to be compared?

include/camp/resource.hpp
```diff
@@ -94,12 +94,20 @@ namespace resources
       void wait_for(Event *e) { m_value->wait_for(e); }
       void wait() { m_value->wait(); }
 
+      bool compare (camp::resources::Resource& r) {
```
MrBurmark, line 97: Everything should only require const references in comparison.

test/resource.cpp
```diff
@@ -51,6 +51,48 @@ TEST(CampResource, GetPlatform)
   ASSERT_EQ(static_cast<const Resource>(Omp()).get_platform(), Platform::omp_target);
 #endif
 }
+TEST(CampResource, Compare)
+{
+  Resource h1{Host()};
+  Resource h2{Host()};
+
+  ASSERT_TRUE(h1 == h1);
```
MrBurmark, line 59: != should be available as well.

include/camp/resource/cuda.hpp
kab163: I think recent changes addressed this, but let me know if something is still missing

include/camp/resource.hpp
kab163: I think Mike's changes addressed this as there is no longer a floating operator overload there, it's been pulled into the class instead... thoughts?

include/camp/resource/host.hpp
```diff
@@ -68,6 +68,25 @@ namespace resources
...
+        return true;
+      }
+      /*
+       * \brief Compares two (Host) resources to see if they are NOT equal.
+       *
+       * \return Always return false. Host resources are always the same.
+       */
+      bool operator!=(Host const&)
+      {
+        return true;
```
mdavis36, line 88: This should probably return false.

include/camp/resource/cuda.hpp
```diff
@@ -256,9 +257,28 @@ namespace resources
         }
       }
 
-      cudaStream_t get_stream() { return stream; }
+      cudaStream_t get_stream() const { return stream; }
       int get_device() { return device; }
```
MrBurmark, line 261: might as well make this const while you're here

include/camp/resource/cuda.hpp
```diff
@@ -256,9 +257,28 @@ namespace resources
...
 
+      /*
+       * \brief Compares two (Cuda) resources to see if they are equal.
+       *
+       * \return True or false depending on if it is the same stream.
+       */
+      bool operator==(Cuda const& c)
```
MrBurmark, line 268: This method should be const

include/camp/resource/host.hpp
MrBurmark, line 88:
```suggestion
        return false;
```

kab163: thanks lol

include/camp/resource.hpp
```diff
@@ -94,12 +94,36 @@ namespace resources
...
+      }
+
     private:
       class ContextInterface
       {
       public:
         virtual ~ContextInterface() {}
         virtual Platform get_platform() const = 0;
+        virtual bool operator==(Resource const& r) = 0;
+        virtual bool operator!=(Resource const& r) = 0;
```
MrBurmark, line 126: It is best practice to use a virtual operator== or make a named method?

mdavis36: Good question, I initially thought using `operator==` here would be best but since this is for a nested class that isn't interfacing with the users I think either would be fine. We don't actually need `operator!=` here so a named method might actually be cleaner in this one case.

include/camp/resource/cuda.hpp
```diff
@@ -29,6 +29,7 @@ namespace resources
   inline namespace v1
   {
     class Cuda;
+    class Resource;
```
kab163, line 32: This isn't needed I don't think...

mdavis36: nope you should be able to remove it

include/camp/resource.hpp
rhornung67: I agree with @mdavis36 on this

include/camp/resource.hpp
```diff
@@ -94,12 +94,36 @@ namespace resources
...
+       * \return True or false depending on comparison with m_value if they are on the same platform.
+       */
+      bool operator==(Resource const& r)
+      {
+        if(get_platform() == r.get_platform()) {
+          return (*m_value == r);
+        }
+        return false;
+      }
+      /*
```
rhornung67, line 109:
```suggestion
      
      /*
```

White space between methods that are not small, one-liners is good.

include/camp/resource.hpp
```diff
@@ -112,11 +136,13 @@ namespace resources
       };
 
       template <typename T>
-      class ContextModel : public ContextInterface
+      class ContextModel final : public ContextInterface
       {
       public:
         ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
         Platform get_platform() const override { return m_modelVal.get_platform(); }
+        bool operator==(Resource const& r) override { return m_modelVal == r.get<T>(); }
```
rhornung67, line 144: Adding a blank line to separate related methods from others helps readability. For example, `==` and `!=` are related, `allocate`, `calloc`, and `deallocate` are related, etc.

include/camp/resource/cuda.hpp
```diff
@@ -256,8 +257,27 @@ namespace resources
...
+      /*
+       * \brief Compares two (Cuda) resources to see if they are equal.
+       *
+       * \return True or false depending on if it is the same stream.
+       */
+      bool operator==(Cuda const& c) const
+      {
+        return (get_stream() == c.get_stream());
+      }
+      /*
```
rhornung67, line 272:
```suggestion
      
      /*
```

Blank line please.

include/camp/resource.hpp
```diff
@@ -94,12 +94,36 @@ namespace resources
       void wait_for(Event *e) { m_value->wait_for(e); }
       void wait() { m_value->wait(); }
 
+      /*
+       * \brief Compares two Resources to see if they are equal.
+       *
+       * \return True or false depending on comparison with m_value if they are on the same platform.
```
rhornung67, line 100: Better to describe the semantics of the comparison (i.e. what does equality mean) vs. the impl details?

include/camp/resource/hip.hpp
```diff
@@ -257,8 +257,27 @@ namespace resources
...
+      /*
+       * \brief Compares two (Hip) resources to see if they are equal.
+       *
+       * \return True or false depending on if this is the same stream.
+       */
+      bool operator==(Hip const& h) const
+      {
+        return (get_stream() == h.get_stream());
+      }
+      /*
```
rhornung67, line 272:
```suggestion
      
      /*
```

include/camp/resource/hip.hpp
```diff
@@ -257,8 +257,27 @@ namespace resources
...
+       * \brief Compares two (Hip) resources to see if they are NOT equal.
+       *
+       * \return True or false depending on result of == operator.
```
rhornung67, line 275:
```suggestion
       * \return Negation of == operator 
```

include/camp/resource/cuda.hpp
```diff
@@ -256,8 +257,27 @@ namespace resources
...
+       * \brief Compares two (Cuda) resources to see if they are NOT equal.
+       *
+       * \return True or false depending on result of == operator.
```
rhornung67, line 275:
```suggestion
       * \return Negation of == operator
```

include/camp/resource/host.hpp
```diff
@@ -68,6 +68,25 @@ namespace resources
...
+      /*
+       * \brief Compares two (Host) resources to see if they are equal.
+       *
+       * \return Always return true since we are on the Host in this case.
+       */
+      bool operator==(Host const&)
+      {
+        return true;
+      }
+      /*
```
rhornung67, line 81:
```suggestion
      
      /*
```

include/camp/resource/omp_target.hpp
```diff
@@ -192,6 +192,24 @@ namespace resources
...
+      /*
+       * \brief Compares two (Omp) resources to see if they are equal.
+       *
+       * \return True or false depending on if this is the same dev int and addr ptr.
+       */
+      bool operator==(Omp const& o)
+      {
+        return (dev == o.dev && addr == o.addr);
+      }
+      /*
```
rhornung67, line 204:
```suggestion
      
      /*
```

include/camp/resource/omp_target.hpp
```diff
@@ -192,6 +192,24 @@ namespace resources
...
+       * \brief Compares two (Omp) resources to see if they are NOT equal.
+       *
+       * \return True or false depending on result of == operator.
```
rhornung67, line 207:
```suggestion
       * \return Negation of == operator
```

include/camp/resource/sycl.hpp
```diff
@@ -199,6 +199,25 @@ namespace resources
...
+      /*
+       * \brief Compares two (Sycl) resources to see if they are equal.
+       *
+       * \return True or false depending on if this is the same queue.
+       */
+      bool operator==(Sycl const& s)
+      {
+        return (get_queue() == s.get_queue());
+      }
+      /*
```
rhornung67, line 211:
```suggestion
      
      /*
```

include/camp/resource/sycl.hpp
```diff
@@ -199,6 +199,25 @@ namespace resources
...
+       * \brief Compares two (Sycl) resources to see if they are NOT equal.
+       *
+       * \return True or false depending on the result of the == operator.
```
rhornung67, line 214:
```suggestion
       * \return Negation of == operator
```

include/camp/resource.hpp
kab163: Making these named methods makes other things freak out... so, still working on this one.

include/camp/resource.hpp
```diff
@@ -94,29 +94,65 @@ namespace resources
...
+      /*
+       * \brief Compares two Resources to see if they are equal. Two Resources are equal if they are
+       * on the same platform and they describe the same stream (i.e. CUDA, HIP) or queue (i.e. SYCL).
+       * (Note: This operator overload is on the generic Resource. Using the ContextModel's operator
+       * overload, we can call the specific resource's (i.e. T) operator== to compare streams or
+       * queues.)
+       *
+       * \return True or false depending on comparison with m_value if they are on the same platform.
+       */
+      bool operator==(Resource const& r)
```
mdavis36, line 106:
```suggestion
      bool operator==(Resource const& r) const
```

kab163: Ah yes, almost forgot. Good catch!

include/camp/resource/sycl.hpp
```diff
@@ -199,6 +199,26 @@ namespace resources
 
       sycl::queue *get_queue() { return qu; }
```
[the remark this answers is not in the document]
kab163: Good catch!

include/camp/resource/sycl.hpp
```diff
@@ -197,7 +197,27 @@ namespace resources
         }
       }
 
-      sycl::queue *get_queue() { return qu; }
+      sycl::queue *get_queue() const { return qu; }
```
trws, line 200: Thanks for adding const in a number of appropriate places, but this one needs a bit of repair.  It needs to have both variants and return a const pointer when const.
```suggestion
      sycl::queue *get_queue() { return qu; }
      sycl::queue const *get_queue() const { return qu; }
```

MrBurmark (dismissed):

trws (changes requested):
I'm pretty happy with this result, but I don't see anything to compare an abstract resource with a concretely typed resource, would that be feasible to add? I think a templated operator == on the Resource class that accepts types with a `get_platform` method would probably do it.

include/camp/resource/sycl.hpp
kab163: Sounds good!

kab163:
> I'm pretty happy with this result, but I don't see anything to compare an abstract resource with a concretely typed resource, would that be feasible to add? I think a templated operator == on the Resource class that accepts types with a `get_platform` method would probably do it.

I guess a comparison between an abstract and concrete resource would be true if they have the same platform? And when would you need to use that (like how can I test it in the test/resource.cpp file)?

include/camp/resource.hpp
```diff
@@ -94,29 +94,74 @@ namespace resources
...
+      }
+
+      template <typename T>
+      bool operator==(T const& t) const
+      {
+        if(get_platform() == t.get_platform()) { 
+          return get<T>().compare(t);
+        }
+        return false;
+      }
```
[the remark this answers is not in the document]
MrBurmark: That's the idea.
The concrete resources are compared with == and not compare, right?
With testing you bring up a good point, all the comparison tests are for type erased resources. Tests comparing concrete resources, comparing type erased and non type erased resources, and comparing non type erased and type erased resources should be added for completeness.
You can make a Resource from a concrete resource by constructing it `Host ch; Resource eh(ch);`, this is what the test is doing with `Resource h1{Host()}`.

include/camp/resource.hpp
```diff
@@ -94,29 +94,74 @@ namespace resources
...
+       * are not equal.
+       *
+       * \return Negation of == operator. 
+       */
+      bool operator!=(Resource const& r) const
+      {
+        return !(*this == r);
+      }
+
+      template <typename T>
```
MrBurmark, line 126: You'll have to disable this method when T is Resource to avoid conflicting with the other `operator==`.
This also only allows you to compare with a concrete resource on the right hand side of an ==, to allow comparison with a concrete resource on the left hand side you'll have to add a namespace scope operator== between T and Resource.
```suggestion
      template <typename T, std::enable_if_t<std::is_same<T, Resource>::value>* = nullptr>
```

trws: Wait, that looks backwards @MrBurmark, it needs to be enabled when T is _not_ same right?

include/camp/resource.hpp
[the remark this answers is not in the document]
MrBurmark: It looks like the Resource constructor is not explicit, so the test may be creating temporary Resource objects instead of doing what we expect.

include/camp/resource.hpp
kab163: thank you both!

test/resource.cpp
```diff
@@ -51,6 +51,67 @@ TEST(CampResource, GetPlatform)
...
+  ASSERT_TRUE(c1 == c1);
+  ASSERT_TRUE(c2 == c2);
+  ASSERT_TRUE(c3 == c);
+  ASSERT_TRUE(c1 != c2);
+  ASSERT_TRUE(c1 != h1);
+
+  ASSERT_FALSE(c1 == c2);
+  ASSERT_FALSE(c2 == c1);
+  ASSERT_FALSE(c1 == h1);
+  ASSERT_FALSE(c1 != c1);
```
MrBurmark, line 80: I'm seeing a lot of repetition, templates to the rescue?

test/resource.cpp
```diff
@@ -51,6 +51,67 @@ TEST(CampResource, GetPlatform)
...
 }
+TEST(CampResource, Compare)
+{
+  Resource h1{Host()};
+  Resource h2{Host()};
+  Host h; Resource h3{h};
+
+  ASSERT_TRUE(h1 == h1);
+  ASSERT_TRUE(h1 == h2);
+  ASSERT_TRUE(h3 == h);
```
MrBurmark, line 62: Please test both orders.
```suggestion
  ASSERT_TRUE(h3 == h);
  ASSERT_TRUE(h == h3);
```

MrBurmark: and both orders of !=

test/resource.cpp
kab163: It seemed like templates were a bit tricky to do here, but i did reduce the repetition a bit

include/camp/resource.hpp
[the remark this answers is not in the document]
kab163: tested with an explicit constructor and that worked well, but in trying to minimize impact to user code, we won't merge that in with this PR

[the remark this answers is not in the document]
trws: Ouch, yeah, could you put up at least an issue or better a PR with that patch so we can track it for a future update or decision?

trws (dismissed):

test/resource.cpp
```diff
@@ -51,6 +51,45 @@ TEST(CampResource, GetPlatform)
...
+#endif
+#ifdef CAMP_HAVE_OMP_OFFLOAD
+  Resource r1{Omp()};
+  Resource r2{Omp()};
+  Omp s; Resource r3{s};
+#endif
+
+  ASSERT_TRUE(h1 == h1);
+  ASSERT_TRUE(h1 == h2);
+  ASSERT_TRUE(h3 == h);
```
MrBurmark, line 78: There should be coverage for all the cases between type-erased and typed resources.
```suggestion
  ASSERT_TRUE(h3 == h);
  ASSERT_FALSE(h3 != h);
  ASSERT_TRUE(h == h3);
  ASSERT_FALSE(h != h3);
```

MrBurmark: There should also be coverage for comparing h to h3, h to r3, s to h3 and s to r3.

kab163: Ooo looks like I broke something... stay tuned lol

MrBurmark: Here's a way to wrap up all the comparisons for a pair of resources.
```suggestion
  auto assert_resources_equal = [](auto const& r1, auto const& r2) {
    ASSERT_TRUE(r1 == r2);
    ASSERT_TRUE(r2 == r1);
    ASSERT_FALSE(r1 != r2);
    ASSERT_FALSE(r2 != r1);
  };
  assert_resources_equal(h1, h2);
  assert_resources_equal(h3, h);
```

kab163:
@trws , @mdavis36 , @rhornung67 - Any ideas how to move forward here? The sycl and windows docker tests are complaining about ambiguous overloads.. I get an error message similar to:

`D:\a\1\s\include\camp/resource.hpp(154,1): error C2666: 'camp::resources::v1::operator ==': 3 overloads have similar conversions 
       D:\a\1\s\include\camp/resource/host.hpp(79,12): message : could be 'bool camp::resources::v1::operator ==(const camp::resources::v1::Host &,const camp::resources::v1::Host &)' [found using argument-dependent lookup] 
       D:\a\1\s\include\camp/resource.hpp(202,10): message : or       'bool camp::resources::v1::operator ==<T,0x0>(const T &,const camp::resources::v1::Resource &)' `

include/camp/resource.hpp
```diff
@@ -144,6 +182,37 @@ namespace resources
       std::shared_ptr<ContextInterface> m_value;
     };
 
+    /*
+     * \brief Compares a generic Resource and a concrete Resource to see if they are equal. To find out, we
+     * first check to see if they are on the same platform. Then we call the operator == to compare the 
+     * type of the generic Resource (T) and the concrete one.
+     *
+     * \return True or false depending on comparison with m_value and t, assuming they are on the same platform.
```
rhornung67, line 190: What does `m_value` refer to here?

kab163: I'll change that, thanks

test/resource.cpp
```diff
@@ -21,8 +21,8 @@
 using namespace camp::resources;
 
 // compatible but different resource for conversion test
-struct Host2 : Host {
-};
+//struct Host2 : Host {
```
rhornung67, line 24: If you are going to leave this in and commented out, please add a comment why it is commented out.

test/resource.cpp
```diff
@@ -35,8 +35,8 @@ TEST(CampResource, ConvertFails)
 {
   Resource h1{Host()};
   h1.get<Host>();
-  ASSERT_THROW(h1.get<Host2>(), std::runtime_error);
-  ASSERT_FALSE(h1.try_get<Host2>());
+  //ASSERT_THROW(h1.get<Host2>(), std::runtime_error);
```
rhornung67, line 38: If you are going to leave this in and commented out, please add a comment why it is commented out.

rhornung67 (commented):
A couple of comments about comments. Plus, one earlier comment I made (about `m_value` in `==` operator comment) still to address.  Otherwise, this looks good.

No more comments. 😄

test/resource.cpp
kab163: Well that I am still trying to work out. When Host2 is left in the test, MSVC and sycl compilers think there is an ambiguous operator== overload and it causes docker failures. When it's gone, everything passes. Not sure how to resolve this. I think Host2 should go back in so that the conversion tests can happen - but I need to figure out how to make MSVC and SYCL not complain about my operator== ...

include/camp/resource.hpp
kab163: In order to appease MSVC and SYCL compilers which were confused about by the dummy struct "Host2" created in the test, we decided to get rid of these specific-generic and generic-specific Resource comparisons... they can be added later perhaps...

test/resource.cpp
kab163: This was put back in and the problematic operator== functions removed

test/resource.cpp
kab163: This was put back in and the problematic operator== functions removed

test/resource.cpp
```diff
@@ -51,6 +51,51 @@ TEST(CampResource, GetPlatform)
...
+  Resource r2{Hip()};
+  Hip s; Resource r3{s};
+#endif
+#ifdef CAMP_HAVE_OMP_OFFLOAD
+  Resource r1{Omp()};
+  Resource r2{Omp()};
+  Omp s; Resource r3{s};
+#endif
+
+  ASSERT_TRUE(h1 == h1);
```
MrBurmark, line 76: Add tests comparing concrete resources like h and s.

kab163: That would require an operator overload that could handle a `camp::resources::v1::Cuda != camp::resources::v1::Host` comparison and I currently don't have that functionality... is it possible to really compare a `T t` and `T t2`? That seems strange - or maybe i'm missing something

MrBurmark: Right, how about just comparing like with like?

test/resource.cpp
```diff
@@ -51,6 +51,53 @@ TEST(CampResource, GetPlatform)
...
+  
+  ASSERT_FALSE(h1 != h2);
+
+#if defined(CAMP_HAVE_CUDA) || \
+    defined(CAMP_HAVE_HIP) || \
+    defined(CAMP_HAVE_OMP_OFFLOAD)
+  ASSERT_TRUE(r1 == r1);
+  ASSERT_TRUE(r2 == r2);
+  ASSERT_TRUE(s == s);
+  ASSERT_TRUE(h == h);
```
MrBurmark, line 87: This should be above, and there should be a != test.

trws (approved):

PR #150, opened 2024-04-23
Make Resource Constructor Explicit

kab163:
Currently, the resource test may be creating temporary Resource objects for certain tests instead of doing what we expect - so making the Resource constructor explicit. 

However, this will require a change to the Reassignment test (and any user code which does a reassignment).

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -54,7 +54,7 @@ namespace resources
                 typename = typename std::enable_if<
                     !std::is_same<typename std::decay<T>::type,
                                   Resource>::value>::type>
-      Resource(T &&value)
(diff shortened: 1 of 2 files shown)


kab163:
Once the Resource constructor is explicit, some of the other operator overload functions could be added back in - see [here](https://github.com/LLNL/camp/blob/1fbc3a0e3f3a2916dcb080ea247e68819c575b02/include/camp/resource.hpp#L188-L217)

davidbeckingsale (commented):
As discussed this morning, I think we want to approach this a different way and use enable_if to prevent attempts to construct a Resource from a string/char*

adayton1 (approved):

PR #152, opened 2024-06-07
Add swap method for array class

adayton1:
* Adds swap method for array class and external swap method for two arrays
* Adds test case for swap


include/camp/array.hpp
```diff
@@ -155,6 +155,12 @@ namespace camp {
          }
       }
 
+      CAMP_HOST_DEVICE constexpr void swap(array& a) noexcept(is_nothrow_swappable<T>::value) {
+         for (std::size_t i = 0; i < N; ++i) {
+            safe_swap(elements[i], a[i]);
+         }
```
trws, line 161: Given the conditional noexcept specification, this should probably be exception safe right?

MrBurmark (dismissed):

include/camp/array.hpp
adayton1: Should I make safe_swap conditionally noexcept if the type is move constructible and move assignable without throwing?

adayton1: I just updated safe_swap. Let me know if that's what you had in mind.

trws: That would work, but if you want this to provide the strong exception guarantee when the arrays contain a potentially throwing move and/or copy, it will need a temporary array to use the copy and swap pattern.  Otherwise an exception can leave both arrays in an inconsistent state.  If the move constructor is noexcept then there's no problem, so it might even make sense to make that a requirement for this method to pass sfinae.

adayton1: I made the swap function conditional on the move constructor/assignment operator being noexcept.

trws (approved):
Looks great, thanks @adayton1!

include/camp/array.hpp
```diff
@@ -155,6 +155,16 @@ namespace camp {
          }
       }
 
+      // TODO: Revisit swap implementations to make them match the std library
+      CAMP_HOST_DEVICE constexpr
+      typename std::enable_if<std::is_nothrow_move_constructible<T>::value &&
+                              std::is_nothrow_move_assignable<T>::value>::type
+      swap(array& a) noexcept(std::is_nothrow_swappable<T>::value) {
```
MrBurmark, line 162: is_nothrow_swappable is C++17

adayton1: Right you are. Converting back to a draft until I have time to look at this again.

MrBurmark (approved):

PR #153, opened 2024-06-11
Allow std tuple_size and tuple_element before C++17

MrBurmark:
There's no reason to not supply them before 17.
They can still be used in code for generic tuple-like classes.


adayton1 (approved):
LGTM

rhornung67 (approved):

long58 (approved):

PR #171, opened 2025-04-17
Updating Camp to C++17

kab163:


kab163:
This is a quick and dirty solution to building without warnings for c++17
- [ ] Should I add back support for C++14 in the CMakeList?
- [ ] Is there another preferred way to ignore the return values of those `no-discard` functions? (I want it to be simple since I just need Camp to shut up and let me update Umpire without complaints.)

trws (dismissed):

test/array.cpp
```diff
@@ -401,7 +401,7 @@ CAMP_TEST_BEGIN(array, structured_binding)
 
 CAMP_TEST_BEGIN(array, deduction_guide)
 {
-   camp::array a{-1, 1};
+   camp::array<int, 2> a{{-1, 1}};
```
adayton1, line 404: This change actually defeats the whole purpose of this test case - it is meant to test that CTAD is set up properly.

kab163: Well I got errors trying to build with CUDA on rzansel:
```
/g/g0/belcher6/camp/test/array.cpp:404:13: error: no viable constructor or deduction guide for deduction of template arguments of 'array'
camp::array a{{(-1), 1}}; 
            ^
/g/g0/belcher6/camp/include/camp/array.hpp:41:8: note: candidate template ignored: couldn't infer template argument 'T'
struct array { 
       ^
/g/g0/belcher6/camp/include/camp/array.hpp:274:1: note: candidate template ignored: couldn't infer template argument 'T'
array(T, U ...)->array< T, (1) + sizeof...(U)> ; 
^
/g/g0/belcher6/camp/include/camp/array.hpp:41:8: note: candidate function template not viable: requires 0 arguments, but 1 was provided
struct array { 
```
and that made it build successfully.

I built with `cmake -DENABLE_CUDA=On -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CUDA_ARCHITECTURES=70 ../` on rzansel (clang 16) - is there another way I should be building?

include/camp/defines.hpp
```diff
@@ -196,7 +196,7 @@ CAMP_DLL_EXPORT cudaError_t cudaAssert(cudaError_t code,
 
 #define campHipErrchk(ans) ::camp::hipAssert((ans), #ans, __FILE__, __LINE__)
 
-#define noReturnCampHipErrchk(ans) (void)::camp::hipAssert((ans), #ans, __FILE__, __LINE__)
+#define CampHipErrchkDiscardReturn(ans) (void)::camp::hipAssert((ans), #ans, __FILE__, __LINE__)
```
davidbeckingsale:
```suggestion
#define campHipErrchkDiscardReturn(ans) (void)::camp::hipAssert((ans), #ans, __FILE__, __LINE__)
```

kab163: Beat you to it!

test/array.cpp
kab163: Thankfully LivChat knows a fix :D

kab163: So apparently in C++17, initializer lists are treated differently during template argument deduction. The deduction guide expects individual arguments (e.g., array(-1, 1)), but with { -1, 1 }, the compiler interprets it as a single initializer list, which doesn't match the deduction guide. Since the compiler cannot infer the type T from { -1, 1 } (because initializer lists are not directly compatible with the deduction guide in this case), it also cannot deduce the size of the array (N) because it doesn't have enough information.

So there is no easy fix without changing the deduction guide.. does that sound right?

kab163: It is worth noting that rocm does not get tripped up by this, but nvcc does. (Hence why I was confused and thought it was fixed when it really wasn't...)

adayton1: I think this is actually an nvcc bug. 

Does `camp::array a{{-1, 1}};` work?

kab163: That gives me:
```
/g/g0/belcher6/camp/test/array.cpp(392): error: no instance of constructor "camp::array" matches the argument list
            argument types are: ({...})
```

kab163: I agree that this seems like an issue with nvcc...

adayton1: From what I understand, this isn't actually a std::initializer_list, but aggregate initialization (https://en.cppreference.com/w/cpp/language/aggregate_initialization).

trws: It is not.  That would be `camp::array a{-1, 1}`, with the perens it's an initializer list.

adayton1: That's what I meant (the original code that you just tagged).

kab163: So how should I proceed with the test? Especially since this only effects the cuda build :(

trws: You're 100% right, sorry, I wasn't looking at the original code. It should work either with `array{...}` or `array(...)` but _not_ with `array({...})`.  If nvcc is being cranky maybe use the traditional construction syntax?

trws: It occurs to me to mention, nvcc is why several camp classes have conditional macros around constructors.  We've had issues in the past where it just flat will not work if we tell it to do aggregate initialization, like for the storage base of `tuple` for example.  I really hope that's not what's happening here, but... yeah, nvcc has had issues with aggregate initialization before.

test/array.cpp
```diff
@@ -399,12 +399,16 @@ CAMP_TEST_BEGIN(array, structured_binding)
           a[1] == 3;
 } CAMP_TEST_END(array, structured_binding)
 
+#if defined(CAMP_ENABLE_CUDA) && (CAMP_CXX_STD == 17)
+//If CUDA is enabled and building with c++17, skip test to avoid NVCC error
+#else
 CAMP_TEST_BEGIN(array, deduction_guide)
 {
-   camp::array<int, 2> a{-1, 1};
+   camp::array a{-1, 1};
```
adayton1, line 407: Did you try Tom's suggestion of `camp::array a(-1, 1);`?

kab163: Yes, when I do that I get the error
```
/g/g0/belcher6/camp/test/array.cpp(407): error: too many initializer values
/g/g0/belcher6/camp/test/array.cpp(407): error: no suitable constructor exists to convert from "int" to "camp::array<int, 2UL>"
```

adayton1: OK. I guess we can disable this test if CUDA is enabled.

test/array.cpp
```diff
@@ -399,12 +399,16 @@ CAMP_TEST_BEGIN(array, structured_binding)
           a[1] == 3;
 } CAMP_TEST_END(array, structured_binding)
 
+#if defined(CAMP_ENABLE_CUDA) && (CAMP_CXX_STD == 17)
```
adayton1, line 402: This could probably be changed to `#if !defined(CAMP_ENABLE_CUDA)` with the `#else` removed. We already check if building with c++17 above: `#if defined(__cplusplus) && __cplusplus >= 201703L`

test/array.cpp
kab163: @adayton1, I did some digging around and there was a suggestion to update the cuda version. I updated to CUDA 11.8.0 and used clang/18.1.8 (loaded the clang/18.1.8-cuda-11.8.0-gcc-11.2.1 module) and it worked on rzansel. So - it looks like this is a bug which has been fixed on later cuda versions than I initially tried (I was running on 11.2.0 before).

kab163: I could add a test for this in the azure pipelines... but I noticed that cuda isn't being turned on for the nvcc docker runs. is that done on purpose?

adayton1 (approved):

davidbeckingsale (approved):

test/array.cpp
trws: Excellent.  I'm glad the update worked.  The extra error you posted means it didn't generate the right aggregate initializer like it should have.  Has the min CUDA version for RAJA advanced that far yet? If so, we might be able to get some compile time back on tuples for nvcc builds as well.

> I noticed that cuda isn't being turned on for the nvcc docker runs. is that done on purpose?

No, unless I missed it, and that's horrifying.  Creating an issue about that right now...

trws (approved):
This looks great, thanks @kab163 for getting this done.  I'm going to create a followup issue to go back through and remove the C++14 and old compiler workarounds, and document the new requirement, but since it wasn't documented before apparently I'm not going to hold this back for it.  Good stuff!

PR #174, opened 2025-05-08
feature: method to get reference to tuple elements satisfying type_trait

johnbowen42:
For context, this is a function I wrote for https://github.com/LLNL/RAJA/pull/1799, where we need to easily extract reference to kernel parameters that are reducers.  The interface requires passing a `tuple` typed object that the user desires to extract elements from, as well as a class-template that inherits from `std::true_type` for desired elements. Here are some alternatives I considered:
- Adding ability to return pointer to element, copy of element, or reference to element.  Right now the method only returns references to elements
- Instead of the type-trait interface, passing the template class directly as a parameter to the method.  This has a number of pitfalls, including the case where the desired template class itself is dependent upon other instantiations of template classes for its definition

I am not committed to the name `get_refs_to_elements_by_type_trait` and would like to hear alternative ideas

### DIFF
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -624,60 +624,63 @@ CAMP_HOST_DEVICE constexpr auto tuple_cat_pair(L&& l, R&& r) noexcept
                         camp::idx_seq_from_t<R>{});
 }
 
-template<template <typename> typename TypeTrait,
-         typename LHS,
-         typename... Params>
-CAMP_HOST_DEVICE constexpr
-concepts::enable_if_t<camp::tuple<LHS&, Params...>, TypeTrait<LHS>>
-get_refs_to_elements_by_type_trait_impl_helper(LHS& lhs,
-                            camp::tuple<Params...> RHS)
+namespace detail
 {
-  return camp::tuple_cat_pair(camp::tuple<LHS&>(lhs),
-                              RHS);
-}
-
-template<template <typename> typename TypeTrait,
-         typename LHS,
-         typename... Params>
-CAMP_HOST_DEVICE constexpr
-concepts::enable_if_t<camp::tuple<Params...>, concepts::negate<TypeTrait<LHS>>>
-get_refs_to_elements_by_type_trait_impl_helper(
-    LHS&,
-    camp::tuple<Params...> RHS)
-{
-  return RHS;
-}
-
-template<camp::idx_t param_size,
-         template <typename> typename TypeTrait,
-         typename TupleType>
-CAMP_HOST_DEVICE constexpr camp::tuple<>
-get_refs_to_elements_by_type_trait_impl(TupleType&,
-                                        camp::num<0>)
-{
-  return camp::tuple<> {};
-}
-
-template<camp::idx_t param_size,
-         template <typename> typename TypeTrait,
-         typename TupleType,
-         camp::idx_t idx>
-CAMP_HOST_DEVICE constexpr auto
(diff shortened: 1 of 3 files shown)


trws:
Interesting, not a bad idea. Would you be willing to add a link to where it's used in that PR? I'm curious about the usage pattern, the implementation looks easy to follow, which is great, but we might be able to make it a bit faster on template expansion depending on the usage.

johnbowen42:
> Interesting, not a bad idea. Would you be willing to add a link to where it's used in that PR? I'm curious about the usage pattern, the implementation looks easy to follow, which is great, but we might be able to make it a bit faster on template expansion depending on the usage.

@trws here https://github.com/LLNL/RAJA/blob/c8f02a9f7c1465d7d019176bb631ea5c65b1f9f8/include/RAJA/pattern/params/forall.hpp#L33

I changed my PR to reflect these changes, but this will break CI for the RAJA PR.  Also, we will need to rewrite camps' git history a little bit and bring in this PR prior to the cxx17 change.

include/camp/tuple.hpp
```diff
@@ -624,6 +624,62 @@ CAMP_HOST_DEVICE constexpr auto tuple_cat_pair(L&& l, R&& r) noexcept
...
+{
+  return camp::tuple<> {};
+}
+
+template<camp::idx_t param_size,
+         template <typename> typename TypeTrait,
+         typename TupleType,
+         camp::idx_t idx>
+CAMP_HOST_DEVICE constexpr auto
+get_refs_to_elements_by_type_trait_impl(TupleType& param,
```
MrBurmark, line 666: Do you want to put these impl functions in a detail namespace?

MrBurmark (dismissed):

artv3 (dismissed):
Looks good to me thanks @johnbowen42 !

test/tuple.cpp
```diff
@@ -416,6 +416,28 @@ TEST(CampTaggedTuple, MakeTagged)
...
+
+template<typename T>
+struct IsSearchType : std::false_type {};
+
+template<typename Param>
+struct IsSearchType<SearchType<Param>> : std::true_type {};
+
+TEST(CampFilterByTypeTrait, MakeFilteredTuple)
+{
+  using BaseTupleType = camp::tuple<double, SearchType<int>, int, OtherType<int>, camp::tuple<double>, SearchType<int>>;
```
MrBurmark, line 433: Do you want to add some more tests like empty tuple, one element tuple, search type in the left most, in the right most, etc?

rhornung67 (approved):

MrBurmark (approved):

artv3 (approved):

MrBurmark:
I assume @trws is talking about either reducing the number of function templates involved and/or using a binary tree kind of approach to reduce the number of recursive function calls.

trws:
This is fine as is, I was hoping we could use an iterative rather than recursive approach for it, but what's coming in doesn't allow for an easy way to do that without a bunch of other infrastructure so it probably isn't worth it right now.

johnbowen42:
> This is fine as is, I was hoping we could use an iterative rather than recursive approach for it, but what's coming in doesn't allow for an easy way to do that without a bunch of other infrastructure so it probably isn't worth it right now.

When we move to cxx17 this should be relatively simple with a fold expression, without building up much infrastructure

trws:
I was thinking so too, but it isn't entirely straightforward because there's no good way to have a fold skip elements or filter into an argument list, especially using emulated folds in 14.  Definitely something to think about when we can though.

PR #176, opened 2025-06-12
Add accessors for dev and addr to Omp resource

MrBurmark:
Add accessors for dev and addr to Omp resource so they can be used in RAJA and RAJAPerf. Currently RAJA always uses the default device and ignores the resource.

### DIFF
--- a/include/camp/resource/omp_target.hpp
+++ b/include/camp/resource/omp_target.hpp
@@ -189,6 +189,10 @@ namespace resources
         }
         return ret;
       }
+
+      int get_device() const { return dev; }
+      char* get_depend_location() const { return addr; }
+
       /*
        * \brief Compares two (Omp) resources to see if they are equal.
        *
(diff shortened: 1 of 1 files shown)


include/camp/resource/omp_target.hpp
```diff
@@ -189,6 +189,10 @@ namespace resources
         }
         return ret;
       }
+
+      int get_device() const { return dev; }
```
adayton1, line 193: Can you document these functions?

adayton1 (approved):

include/camp/resource/omp_target.hpp
kab163: +1 on documentation :D

kab163: If you give me descriptions, I can help with the PR. The more docs the better :D

MrBurmark: I can document them. They are equivalent to the get_device and get_stream methods on the Cuda and Hip resources.

kab163 (approved):
Thanks!

adayton1 (approved):

PR #177, opened 2025-06-12
Add tests for Sycl resource and more for Omp resource

MrBurmark:
Generalize resource tests and test all resource types.

### DIFF
--- a/test/resource.cpp
+++ b/test/resource.cpp
@@ -333,6 +333,22 @@ TEST(CampResource, GetEvent)
     campHipErrchkDiscardReturn(hipStreamDestroy(s));
   }
 #endif
+#ifdef CAMP_HAVE_OMP_OFFLOAD
+  {
+    char a[1];
+    test_get_event<Omp, OmpEvent>(&a[0]);
+  }
+#endif
+#ifdef CAMP_HAVE_SYCL
+  {
+    auto gpuSelector = sycl::gpu_selector_v;
+    sycl::property_list propertyList =
+        sycl::property_list(sycl::property::queue::in_order());
+    sycl::context context;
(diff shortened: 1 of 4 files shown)


test/resource.cpp
```diff
@@ -13,70 +13,146 @@
...
+  struct Omp2 : Omp { };
+#endif
+#ifdef CAMP_HAVE_SYCL
+  struct Sycl2 : Sycl { };
+#endif
+
+template < typename Res >
+void test_construct()
+{
+  Resource h1{Res()};
```
adayton1, line 33: Maybe use `r` instead of `h1`?

test/resource.cpp
```diff
@@ -13,70 +13,149 @@
...
+  ASSERT_EQ(static_cast<const Resource>(Sycl()).get_platform(), Platform::sycl);
+#endif
+}
 
-  ASSERT_TRUE(h1 == h1);
-  ASSERT_TRUE(h1 == h2);
-  
-  ASSERT_FALSE(h1 != h2);
+template < typename Res >
+void test_compare(Resource& h1, Resource& h2, Resource& h3)
```
adayton1, line 150: This test is hard for me to follow. Is it testing the following cases?
- Self equality (but then why no r3 == r3 or r == r3?)
- Inequality between different objects of the same resource type (but then why no r1 != r3 or r2 != r3 and the reverse ordering?)
- Inequality between different resource types (but then why no r2 != h2 or checks of the reverse ordering?)

MrBurmark: Its probably also helpful to know that all Host resources compare == to all other Host resources.

adayton1 (dismissed):

PR #179, opened 2025-06-19
bug: Fix failing sycl test and fix Docker to  run test

johnbowen42:
resource.offload breaks RAJA sycl builds https://github.com/LLNL/RAJA/actions/runs/15746086961/job/44382460013?pr=1859.  This test is unfortunately not currently built or run as part of the Docker sycl pipeline either.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -96,6 +96,7 @@ jobs:
         compiler: /opt/rocm/llvm/bin/amdclang++
       sycl:
         base_img: oneapi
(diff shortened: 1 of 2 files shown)


rhornung67 (approved):
@johnbowen42 we don't have access to GPU resources on azure. Can SYCL tests be run there?

johnbowen42:
> @johnbowen42 we don't have access to GPU resources on azure. Can SYCL tests be run there?

I guess not. I am validating this change in RAJA CI before merging

PR #194, opened 2026-03-04
add tuple_size to camp::array

gberg617:
Add tuple_size to camp::array.  This is needed in order to use camp::apply on arrays.

### DIFF
--- a/include/camp/array.hpp
+++ b/include/camp/array.hpp
@@ -31,7 +31,6 @@
 
 namespace camp
 {
-
 ///
 /// Provides a portable std::array-like class.
 ///
--- a/include/camp/array.hpp
+++ b/include/camp/array.hpp
@@ -22,8 +22,8 @@
 
 #include <cstddef>
 #include <stdexcept>
-#include <tuple>
 #include <type_traits>
+#include <utility>
 
 #include "camp/defines.hpp"
 #include "camp/helpers.hpp"
--- a/include/camp/array.hpp
+++ b/include/camp/array.hpp
@@ -22,6 +22,7 @@
 
 #include <cstddef>
 #include <stdexcept>
+#include <tuple>
 #include <type_traits>
 
 #include "camp/defines.hpp"
@@ -30,8 +31,6 @@
 
 namespace camp
 {
-template <typename Tuple>
-struct tuple_size;
 
 ///
 /// Provides a portable std::array-like class.
@@ -286,10 +285,6 @@ CAMP_HOST_DEVICE constexpr void swap(array<T, N>& a, array<T, N>& b) noexcept(
 {
   a.swap(b);
 }
-
-template <typename T, size_t N>
-struct tuple_size<array<T, N>> : ::camp::num<N> {
-};
 
 namespace detail
 {
--- a/include/camp/array.hpp
+++ b/include/camp/array.hpp
@@ -30,6 +30,9 @@
 
 namespace camp
 {
+template <typename Tuple>
+struct tuple_size;
+
 ///
 /// Provides a portable std::array-like class.
 ///
@@ -283,6 +286,10 @@ CAMP_HOST_DEVICE constexpr void swap(array<T, N>& a, array<T, N>& b) noexcept(
 {
   a.swap(b);
 }
+
+template <typename T, size_t N>
+struct tuple_size<array<T, N>> : ::camp::num<N> {
+};
 
 namespace detail
 {
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -17,6 +17,7 @@
 #include <sstream>
 #include <type_traits>
 
+#include "camp/array.hpp"
 #include "camp/concepts.hpp"
 #include "camp/map.hpp"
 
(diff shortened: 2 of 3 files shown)


rhornung67 (approved):

include/camp/tuple.hpp
```diff
@@ -52,6 +53,10 @@ template <typename... Args>
 struct tuple_size<tuple<Args...>> : ::camp::num<sizeof...(Args)> {
 };
 
+template <typename T, size_t N>
+struct tuple_size<array<T, N>> : ::camp::num<N> {
```
adayton1, line 57: There's already a tuple_size entity in array.hpp. Does that one not work?

gberg617: That one does not work for me because it's defined within the std namespace. `camp::apply` looks for `camp::tuple_size`, so we need to define one that specializes for `camp::array`.

adayton1: Can you move this one to array.hpp, then?

adayton1:
I'm actually wondering if we should remove camp::array since we should be able to use std::array in device code now that we require c++17 (almost everything in std::array is constexpr in c++17, and everything is constexpr in c++20). With nvcc you have to supply the `--expt-relaxed-constexpr` flag, which will make all constexpr function `__host__ __device__`, but with amdclang, constexpr automatically implies `__host__ __device__`.

gberg617:
> I'm actually wondering if we should remove camp::array since we should be able to use std::array in device code now that we require c++17 (almost everything in std::array is constexpr in c++17, and everything is constexpr in c++20). With nvcc you have to supply the `--expt-relaxed-constexpr` flag, which will make all constexpr function `__host__ __device__`, but with amdclang, constexpr automatically implies `__host__ __device__`.

I would be fine with this.  It doesn't look like it's used in RAJA, other than in my SubView PR.

gberg617:
See https://github.com/llnl/camp/pull/195

include/camp/tuple.hpp
rhornung67: @gberg617 can you please make this change and then we will merge this?

gberg617: I made that change.  I ended up also needing to forward-declare `tuple_size` in array.hpp to prevent include-ordering issues.

include/camp/tuple.hpp
```diff
@@ -19,6 +19,7 @@
 #include <sstream>
 #include <type_traits>
 
+#include "camp/array.hpp"
```
adayton1: This still seems backwards. tuple should be able to exist without knowing about array.

adayton1: If array needs tuple_size from tuple.hpp, then just have array.hpp include tuple.hpp. But tuple should be able to stand alone independent of array.

adayton1: Really the base declaration of tuple_size and tuple_element should be in a separate header that both tuple.hpp and array.hpp could include.

MrBurmark: Where is camp apply defined? That's what this is all about, is that in some higher level common header?

MrBurmark: Does it make more sense to change camp::apply to use std::tuple_size?

gberg617: What about two separate headers: one for traits common to both array and tuple (such as tuple_size), and another for common implementations (such as apply, invoke, etc..)?

MrBurmark: tuple_size isn't used in that many places, how about just using std::tuple_size as we should have been doing all along? We can make camp::tuple_size an alias for std::tuple_size.

gberg617: Sure, I think something like that can work. My previous point was about moving camp's `tuple_size` and `apply(...)` to a separate header so that traits/methods that are common to both camp tuples and arrays don't live in tuple.hpp, but are instead included in there as well as array.hpp.  Hopefully this would address the concerns that @adayton1 mentions above.

MrBurmark: I made `camp::tuple_size` and `camp::tuple_element` into aliases for `std::tuple_size` and `std::tuple_element`.
I also added a test that camp::apply works with tuple-like types.

include/camp/array.hpp
```diff
@@ -22,6 +22,7 @@
 
 #include <cstddef>
 #include <stdexcept>
+#include <tuple>
```
adayton1: Can this be removed now?

MrBurmark: I added it to make sure `std::tuple_size` and `std::tuple_element` were declared.

adayton1: Aren't they declared in `<utility>`? Though that's probably not any better than including tuple unless tuple includes utility.

adayton1 (dismissed):

adayton1 (approved):

trws:
This looks like a nice way to trim it down, but does the result work with `tagged_tuple` and other camp types correctly? We had a couple of specializations that aren't reflected after the change.

MrBurmark:
The main difference seems to be the cv qualifier specializations which the std handles in the same way.
See "2-4) Specializations for a cv-qualified types reuse the value from the corresponding cv-unqualified versions by default." in https://en.cppreference.com/cpp/utility/tuple_size

trws:
I missed that the main definitions were still there and only saw the cv version removal, all good.

trws (approved):

PR #195, opened 2026-03-05
remove camp::array

gberg617:
Remove camp::array since it is unused.


MrBurmark:
How about we keep this in until we've truly moved to C++20.
After that do we want to have camp::array be an alias for std::array?
Have people actually tried using std::array in device code? This is a nice workaround in case anything goes wrong.

trws:
If std::array from both libc++ and libstdc++ work in device code I would be fine with that, but last I knew neither of them did. Has that changed?

PR #196, opened 2026-04-15
Feature/smith84/cuda 13 support

smithsg84:
Specialization of camp I/O for CUDA 13 cudaMemLocation.

Invoking CUDA methods via CAMP_CUDA_API_INVOKE_AND_CHECK() may have cudaMemLocation arguments which require I/O for error reporting messages.  In CUDA < 13 these were simple int enum values but in CUDA  13 cudaMemLocation it is a struct.

### DIFF
--- a/include/camp/helpers.hpp
+++ b/include/camp/helpers.hpp
@@ -300,7 +300,7 @@ namespace experimental
   // 13 cudaMemLocation it is a struct.
 
   //! Convert CUDA cudaMemLocationType to string
-  inline const char* to_string(cudaMemLocationType t)
+  constexpr std::string_view to_string(cudaMemLocationType t)
   {
     switch (t) {
       case cudaMemLocationTypeInvalid:
@@ -320,11 +320,11 @@ namespace experimental
   inline std::ostream& print_cudaMemLocation(std::ostream& os,
 					     const cudaMemLocation& loc)
   {
-    os << "{" <<  to_string(loc.type) << "," << loc.id << "}";
+    os << "{" <<  to_string(loc.type) << ", " << loc.id << "}";
     return os;
   }
 
-  //! Specialization for printing of cudaMemLocation&
+  //! Specialization for printing cudaMemLocation&
   //
   // Invoking CUDA methods via CAMP_CUDA_API_INVOKE_AND_CHECK() may
   // have cudaMemLocation arguments which require I/O for error
--- a/include/camp/helpers.hpp
+++ b/include/camp/helpers.hpp
@@ -304,13 +304,13 @@ namespace experimental
   {
     switch (t) {
       case cudaMemLocationTypeInvalid:
-	return "Invalid";
+	return "cudaMemLocationTypeInvalid";
       case cudaMemLocationTypeDevice:
-	return "Device";
+	return "cudaMemLocationTypeDevice";
       case cudaMemLocationTypeHost:
-	return "Host";
+	return "cudaMemLocationTypeHost";
       case cudaMemLocationTypeHostNuma:
-	return "HostNuma";
+	return "cudaMemLocationTypeHostNuma";
       default:
 	return "Unknown";
     }
@@ -320,8 +320,7 @@ namespace experimental
   inline std::ostream& print_cudaMemLocation(std::ostream& os,
 					     const cudaMemLocation& loc)
   {
-    os << "cudaMemLocation{type=" << to_string(loc.type)
-       << ", id=" << loc.id << "}";
+    os << "{" <<  to_string(loc.type) << "," << loc.id << "}";
     return os;
   }
 
--- a/include/camp/helpers.hpp
+++ b/include/camp/helpers.hpp
@@ -288,11 +288,25 @@ namespace experimental
   }
 
 
+
 #ifdef CAMP_ENABLE_CUDA
+  
+// SGS Question should the specialization for CUDA 13 cudaMemLocation printing be here?
+// SGS Question Is camp requiring C++20 as well so use "requires std::is_same_v()" simplify non-const / const specialization?
+//  template <typename T>
+//  struct StreamInsertHelper<T&>
+//  requires std::is_same_v<std::remove_cv_t<T>, cudaMemLocation>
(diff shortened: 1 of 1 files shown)


artv3 (dismissed):

include/camp/helpers.hpp
```diff
@@ -288,7 +288,75 @@ namespace experimental
...
+	return "Unknown";
+    }
+  }
+
+  //! Print helper for cudaMemLocation
+  inline std::ostream& print_cudaMemLocation(std::ostream& os,
+					     const cudaMemLocation& loc)
+  {
+    os << "cudaMemLocation{type=" << to_string(loc.type)
+       << ", id=" << loc.id << "}";
```
MrBurmark: We haven't printed the type name or member names in the StreamInsertHelper specializations before.

include/camp/helpers.hpp
```diff
@@ -288,7 +288,75 @@ namespace experimental
...
+  // cudaMemLocation arguments which require I/O for error reporting
+  // messages.  In CUDA < 13 these were simple int enum values but in CUDA
+  // 13 cudaMemLocation it is a struct.
+
+  //! Convert CUDA cudaMemLocationType to string
+  inline const char* to_string(cudaMemLocationType t)
+  {
+    switch (t) {
+      case cudaMemLocationTypeInvalid:
+	return "Invalid";
```
MrBurmark: Would it be better to print the whole name?
```suggestion
      case cudaMemLocationTypeInvalid:
	return "cudaMemLocationTypeInvalid";
```

include/camp/helpers.hpp
```diff
@@ -288,7 +288,74 @@ namespace experimental
...
+      case cudaMemLocationTypeDevice:
+	return "cudaMemLocationTypeDevice";
+      case cudaMemLocationTypeHost:
+	return "cudaMemLocationTypeHost";
+      case cudaMemLocationTypeHostNuma:
+	return "cudaMemLocationTypeHostNuma";
+      default:
+	return "Unknown";
+    }
+  }
```
artv3:
```suggestion
inline constexpr std::string_view to_string(cudaMemLocationType t)
{
  switch (t) {
    case cudaMemLocationTypeInvalid:
      return "cudaMemLocationTypeInvalid";
    case cudaMemLocationTypeDevice:
      return "cudaMemLocationTypeDevice";
    case cudaMemLocationTypeHost:
      return "cudaMemLocationTypeHost";
    case cudaMemLocationTypeHostNuma:
      return "cudaMemLocationTypeHostNuma";
    default:
      return "Unknown";
  }
}
```

And there would need to be the following header: #include <string_view>

artv3: My suggestion if we want to move towards more modern C++ and if we need the read-only string capability

include/camp/helpers.hpp
```diff
@@ -288,7 +288,74 @@ namespace experimental
...
+
+  //! Print helper for cudaMemLocation
+  inline std::ostream& print_cudaMemLocation(std::ostream& os,
+					     const cudaMemLocation& loc)
+  {
+    os << "{" <<  to_string(loc.type) << "," << loc.id << "}";
```
artv3:
```suggestion
    os << "{" <<  to_string(loc.type) << ", " << loc.id << "}";
```
having a space might improve readability

include/camp/helpers.hpp
```diff
@@ -288,7 +288,74 @@ namespace experimental
...
+    return os;
+  }
+
+  //! Specialization for printing of cudaMemLocation&
```
artv3:
```suggestion
  //! Specialization for printing cudaMemLocation&
```
nitpicking...

include/camp/helpers.hpp
```diff
@@ -288,7 +288,74 @@ namespace experimental
...
+  //
+  // Invoking CUDA methods via CAMP_CUDA_API_INVOKE_AND_CHECK() may
+  // have cudaMemLocation arguments which require I/O for error
+  // reporting messages.  CUDA does not supply an operator<< for
+  // cudaMemLocation.  Specialize the StreamInsertHelper rather than
+  // operator<< to avoid conflicts.
+  template <>
+  struct StreamInsertHelper<cudaMemLocation&> {
+    cudaMemLocation& m_val;
```
artv3:
```suggestion
    const cudaMemLocation& m_val;
```
Do we expect to modify `m_val` ? if not this becomes very similar to the Specialization for printing of const cudaMemLocation& right below

MrBurmark: The two specializations are unfortunate but required to get all the cases so they don't go to the unspecialized StreamInsertHelper.

smithsg84: > The two specializations are unfortunate but required to get all the cases so they don't go to the unspecialized StreamInsertHelper.

For C++ 20 I've seen the non const / const done using requires.   Is this a pattern we should move to to reduce duplication?   Is Camp moving to C++20?

```
template <class T>
  requires std::same_as<std::remove_cv_t<T>, cudaMemLocation>
struct StreamInsertHelper<T&> {
  T& m_val;

  std::ostream& operator()(std::ostream& str) const {
    return print_cudaMemLocation(str, m_val);
  }
};`
```

artv3: I think @johnbowen42 was working on supporting C++20 in camp

MrBurmark: @smithsg84 Let's leave it as-is for now, but make a note/issue to simplify this as part of the c++20 changes.

MrBurmark (dismissed):

rhornung67 (dismissed):

include/camp/helpers.hpp
artv3: @smithsg84 , Do we need the member to be modifiable or can it be marked const? A bit nitpicky here.... but just to close the loop on this.

artv3 (approved):

rhornung67 (approved):

trws (approved):

MrBurmark (approved):

PR #197, opened 2026-04-20
feature: Bump camp to C++20, use C++20 concepts

johnbowen42:
This PR changes camp's CMake to require C++20, and modifies concepts.hpp to use C++20 concepts instead of camp's pre-C++20 concept interface, which generated SFINAE structs.


include/camp/concepts.hpp
```diff
@@ -172,112 +185,82 @@ namespace concepts
...
-                      convertible_to<bool>(val<U>() == val<T>()),
-                      convertible_to<bool>(val<T>() == val<U>()),
-                      convertible_to<bool>(val<U>() != val<T>()),
-                      convertible_to<bool>(val<T>() != val<U>())){};
-
-  template <typename T>
-  struct Comparable : ComparableTo<T, T> {
+  concept ComparableTo = requires (T lhs, U rhs) {
+        {lhs < val<T>()} -> std::convertible_to<bool>;
+        {lhs < val<U>()} -> std::convertible_to<bool>;
```
MrBurmark, line 198:
```suggestion
        {lhs < lhs} -> std::convertible_to<bool>;
        {lhs < rhs} -> std::convertible_to<bool>;
```

include/camp/concepts.hpp
```diff
@@ -172,112 +185,82 @@ namespace concepts
...
-  template <typename T>
-  struct Range
-      : DefineConcept(HasBeginEnd<T>(), Iterator<iterator_from<T>>()){};
+  concept Comparable = ComparableTo<T, T>;
+
+  DefineConceptBool(Arithmetic, std::is_arithmetic<T>())
+  DefineConceptBool(FloatingPoint, std::is_floating_point<T>())
+  DefineConceptBool(Integral, std::is_integral<T>())
+  DefineConceptBool(Signed, std::is_signed<T>())
+  DefineConceptBool(Unsigned, std::is_unsigned<T>())
```
MrBurmark, line 218: Should we use the std concept instead of the std type_trait?

include/camp/concepts.hpp
rhornung67: @MrBurmark the conditional in your suggested change at line 197 will never be true.

MrBurmark: But it will compile, which is what is required here.

include/camp/concepts.hpp
artv3: yes

include/camp/concepts.hpp
bechols97: There is a `std::totally_ordered_with<T, U>` concept from the std concepts header that seems to do similar checks as well. Would it make sense to use the std concepts version here as well?

johnbowen42: I think technically that would impose extra constraints vs what was here before, namely 
```
1) std::totally_ordered<T> is modeled only if, given lvalues a, b and c of type const std::remove_reference_t<T>:

    Exactly one of bool(a < b), bool(a > b) and bool(a == b) is true;
    If bool(a < b) and bool(b < c) are both true, then bool(a < c) is true;
    bool(a > b) == bool(b < a)
    bool(a >= b) == !bool(a < b)
    bool(a <= b) == !bool(b < a)
```
the semantic requirements here https://en.cppreference.com/cpp/concepts/totally_ordered have more detail

include/camp/concepts.hpp
johnbowen42: thanks for pointing this out

include/camp/concepts.hpp
MrBurmark: Ya, I think the totally_ordered concept has the same requirements in the actual requires clause, but has extra requirements that can't be tested with a requires clause as @johnbowen42 pointed out in its semantic requirements.

include/camp/concepts.hpp
```diff
@@ -172,112 +186,82 @@ namespace concepts
...
+        // Reference increment requirements
+        {arg_ref += increment} -> std::same_as<T&>;
+        {arg_ref -= increment} ->std::same_as<T&>;
+        // Value increment requirements
+        {arg + increment} -> std::same_as<T>;
+        {increment + arg} -> std::same_as<T>;
+        {arg - increment} -> std::same_as<T>;
+        // random access index operator
+        arg[increment];
+    };
```
MrBurmark: Are there not std concepts for these?

johnbowen42: there are but the most basic iterator requirements are a little more restrictive than ours https://en.cppreference.com/cpp/header/iterator

johnbowen42: I will look into this more and try to figure out if we can substitute them

MrBurmark: How about we leave a comment or make an issue of which std concepts are more restrictive than ours, then we can use ours for now and move to the std one later?

include/camp/concepts.hpp
```diff
@@ -67,14 +68,27 @@ namespace concepts
...
+concept CONCEPT_NAME = EXPAND_AND(__VA_ARGS__);
 
 #define DefineTypeTraitFromConcept(TTName, ConceptName)             \
-  template <typename... Args>                                       \
-  struct TTName : camp::concepts::requires_<ConceptName, Args...> { \
-  }
+  template<class T>                                                 \
+  struct TTName : std::bool_constant<ConceptName<T>> {};            \
+  template<class T>                                                 \
+  inline constexpr bool TTName##_v = is_iterator<T>::value;
```
artv3:
```suggestion
#define DefineTypeTraitFromConcept(TTName, ConceptName)   \
  template<class T>                                       \
  struct TTName : std::bool_constant<ConceptName<T>> {};  \
  template<class T>                                       \
  inline constexpr bool TTName##_v = TTName<T>::value;
```

With the way you have it defined doesn't it mean that every _v variable generated by the macro becomes tied to is_iterator<T>::value, regardless of the actual trait?

include/camp/concepts.hpp
```diff
@@ -172,112 +186,82 @@ namespace concepts
...
-      : DefineConcept(HasBeginEnd<T>(), Iterator<iterator_from<T>>()){};
-
-  template <typename T>
-  struct ForwardRange
-      : DefineConcept(HasBeginEnd<T>(), ForwardIterator<iterator_from<T>>()){};
+  concept Comparable = ComparableTo<T, T>;
+
+  DefineConceptFromSTL(FloatingPoint, std::floating_point<T>)
+  DefineConceptFromSTL(Integral, std::integral<T>)
+  DefineConceptFromSTL(Arithmetic, FloatingPoint<T> || Integral<T>)
```
artv3: I don't think Arithmetic falls from STL

include/camp/concepts.hpp
```diff
@@ -299,14 +283,18 @@ namespace type_traits
...
-  DefineTypeTraitFromConcept(is_comparable_to, camp::concepts::ComparableTo);
 
   DefineTypeTraitFromConcept(is_arithmetic, camp::concepts::Arithmetic);
   DefineTypeTraitFromConcept(is_floating_point, camp::concepts::FloatingPoint);
   DefineTypeTraitFromConcept(is_integral, camp::concepts::Integral);
   DefineTypeTraitFromConcept(is_signed, camp::concepts::Signed);
   DefineTypeTraitFromConcept(is_unsigned, camp::concepts::Unsigned);
 
+  template<class T>
+  struct is_comparable_to : std::bool_constant<camp::concepts::ComparableTo<T, T>> {};
```
artv3: This looks like a self comparison, not a `to` comparison`

include/camp/concepts.hpp
```diff
@@ -172,112 +186,82 @@ namespace concepts
...
-      : DefineConcept(HasBeginEnd<T>(), ForwardIterator<iterator_from<T>>()){};
+  concept Comparable = ComparableTo<T, T>;
+
+  DefineConceptFromSTL(FloatingPoint, std::floating_point<T>)
+  DefineConceptFromSTL(Integral, std::integral<T>)
+  DefineConceptFromSTL(Arithmetic, FloatingPoint<T> || Integral<T>)
+  DefineConceptFromSTL(Signed, std::signed_integral<T>)
+  DefineConceptFromSTL(Unsigned, std::unsigned_integral<T>)
+
+  DefineConceptBinary(Iterator, !Integral<T>, requires(T arg, T& arg_ref) {
```
artv3: Why not support Integral?

include/camp/concepts.hpp
```diff
@@ -172,112 +186,82 @@ namespace concepts
...
-      : DefineConcept(convertible_to<bool>(val<T>() <= val<T>())){};
-
-  template <typename T>
-  struct GreaterEqualComparable
-      : DefineConcept(convertible_to<bool>(val<T>() >= val<T>())){};
-
-  template <typename T>
-  struct EqualityComparable
-      : DefineConcept(convertible_to<bool>(val<T>() == val<T>())){};
+  DefineConceptRequires(Swappable, swap(arg, arg);)
```
artv3: Maybe use  std::swappable<T>  instead?

include/camp/concepts.hpp
```diff
@@ -172,112 +186,82 @@ namespace concepts
...
+        {arg - increment} -> std::same_as<T>;
+        // random access index operator
+        arg[increment];
+    };
 
   template <typename T>
-  struct BidirectionalRange
-      : DefineConcept(HasBeginEnd<T>(),
-                      BidirectionalIterator<iterator_from<T>>()){};
+  concept HasBeginEnd = requires(T arg) {
```
artv3: This looks similar to this: https://en.cppreference.com/cpp/ranges/range can that be used at some capacity?

include/camp/concepts.hpp
MrBurmark: integers aren't iterators

artv3: right so it seems unnecessary? I guess its there for explicitness?

johnbowen42:
note on this PR:  camp's CI is a little old and the existing images aren't compatible with C++20 compilers and headers.  I am currently updating and migrating camp to use radiuss-docker: https://github.com/llnl/radiuss-docker/pkgs/container/radiuss/versions.  I will address everybody's comments once I figure out CI stuff

include/camp/concepts.hpp
johnbowen42: I'll make a note of this: for example https://en.cppreference.com/cpp/iterator/weakly_incrementable is a concept which would be useful, but requires both pre and post increment

include/camp/concepts.hpp
trws: It can be useful to avoid implicit conversion from integral to pointer, which is an "iterator" as far as concepts are concerned.

include/camp/concepts.hpp
trws: That should work, just make sure there isn't an implicit requirement that the end return a sentinel.  I don't _think_ there is.  Also we should be using ranges begin/end _everywhere_ because it doesn't require the stupid "using" dance we have to do to get the default overloads scoped for those.

include/camp/concepts.hpp
artv3: ah interesting, thanks

include/camp/concepts.hpp
johnbowen42: I left comments explaining this--let me know if you think anything can be clearer

include/camp/concepts.hpp
johnbowen42: It looks like the `ranges::end` concept does require a sentinel.  Should we try to move towards using `std::ranges::begin` and `std:ranges::end` given this?

include/camp/concepts.hpp
johnbowen42: thanks Arturo, fixed

include/camp/concepts.hpp
```diff
@@ -67,14 +68,27 @@ namespace concepts
...
+template<typename T> \
+concept CONCEPT_NAME = requires(T arg) { expr };
+
+#define DefineConceptFromSTL(CONCEPT_NAME, stl_concept) \
+template<typename T> \
+concept CONCEPT_NAME = stl_concept;
+
+#define DefineConceptBinary(CONCEPT_NAME, expr1, expr2) \
+template<typename T> \
+concept CONCEPT_NAME = expr1 && expr2;
```
bechols97: nit: Since this is always using the logical and operator, should this have a more descriptive name?

include/camp/concepts.hpp
artv3: Just wanted to follow up here.

include/camp/concepts.hpp
artv3: yes can we try?

include/camp/concepts.hpp
johnbowen42: although `std::is_arithmetic` is a type traits, there is no `std::arithmetic` concept, just `std::floating_point`, `std::integral`, and signed/unsigned integral concepts. Should I just define our `Arithmetic` concept from the STL type_trait, so that its fully backwards compatible?

artv3 (approved):

bechols97 (approved):

Issue #199, opened 2026-05-21
Apparent incorrect definition of `prepend`


dkachuma:
Looking at [camp.hpp:45-48](https://github.com/llnl/camp/blob/main/include/camp/camp.hpp#L45-L48), I'm surprised that `prepend` looks like `append`. I was expecting
```cpp
template <typename... Elements, typename T>
struct prepend<list<Elements...>, T> {
  using type = list<T, Elements...>;
};
```

trws:
Wow, that's clearly a mistake on my part.  Would you like to put that up as a PR so you get credit for it?

dkachuma:
> Wow, that's clearly a mistake on my part. Would you like to put that up as a PR so you get credit for it?

I can create a fork. Not sure if you're able to merge PRs from forks.

PR #200, opened 2026-05-22
Fix prepend

dkachuma:
Bug Fix in prepend (`include/camp/camp.hpp`):
Corrects the logic for the prepend metaprogramming function. Previously, it was incorrectly appending the type T to the end of the list (`list<Elements..., T>`). The patch fixes this so it properly prepends T to the front of the list (`list<T, Elements...>`).

Added unit tests for `prepend`, `append` and `extend`.

#199


rhornung67:
@dkachuma thank you for the contribution. I pulled the PR branch from your fork into the camp repo for review. It is here: https://github.com/llnl/camp/pull/201

PR #203, opened 2026-05-26
Implement SyclEvent

MrBurmark:
Add an implementation for `SyclEvent`.

This gets an event from the queue by calling submit but not actually submitting any work. Hopefully this gets an event without causing any significant overheads and is well defined behavior.

Note that this will not work for out of order queues. So now the `SyclEvent` constructor throws an exception if you try to create an event from an out of order queue. We only make in order queues due to our semantic guarantees, but someone could make a `Sycl` resource via `SyclFromQueue` from an out of order queue. 

Depends on #182 .

### DIFF
--- a/include/camp/resource/sycl.hpp
+++ b/include/camp/resource/sycl.hpp
@@ -34,16 +34,28 @@ namespace resources
     class SyclEvent
     {
     public:
-      // TODO: make this actually work
-      SyclEvent(sycl::queue& CAMP_UNUSED_ARG(qu)) { m_event = sycl::event(); }
+      explicit SyclEvent(sycl::event e)
+        : m_event(std::move(e))
+      {}
+
+      // TODO: see what overhead an empty submit has
+      SyclEvent(sycl::queue& qu)
+        : m_event(qu.submit([&](::sycl::handler& CAMP_UNUSED_ARG(h)) {}))
+      {}
 
       SyclEvent(Sycl& res);
 
-      bool check() const { return true; }
+      bool check() const
+      {
+        return m_event.get_info<sycl::info::event::command_execution_status>()
+            == sycl::info::event_command_status::complete;
+      }
 
-      void wait() const { getSyclEvent_t().wait(); }
+      void wait() const { m_event.wait(); }
 
-      sycl::event getSyclEvent_t() const { return m_event; }
+      sycl::event& getSyclEvent_t() { return m_event; }
+
+      sycl::event const& getSyclEvent_t() const { return m_event; }
 
     private:
       sycl::event m_event;
@@ -356,7 +368,7 @@ namespace resources
 
     inline SyclEvent::SyclEvent(Sycl &res)
       : SyclEvent(res.get_queue())
-    { }
+    {}
 
   }  // namespace v1
 
--- a/include/camp/resource/sycl.hpp
+++ b/include/camp/resource/sycl.hpp
@@ -40,8 +40,12 @@ namespace resources
 
       // TODO: see what overhead an empty submit has
       SyclEvent(sycl::queue& qu)
-        : m_event(qu.submit([&](::sycl::handler& CAMP_UNUSED_ARG(h)) {}))
-      {}
+      {
+        if (!qu.is_in_order()) {
+          ::camp::throw_re("Queue is not in_order.");
+        }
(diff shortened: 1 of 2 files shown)


MrBurmark:
@trws does this seem like a reasonable way to implement `SyclEvent`?

adayton1 (dismissed):

adayton1 (dismissed):

include/camp/resource/sycl.hpp
```diff
@@ -34,16 +34,28 @@ namespace resources
     class SyclEvent
     {
     public:
-      // TODO: make this actually work
-      SyclEvent(sycl::queue& CAMP_UNUSED_ARG(qu)) { m_event = sycl::event(); }
+      explicit SyclEvent(sycl::event e)
+        : m_event(std::move(e))
+      {}
+
+      // TODO: see what overhead an empty submit has
```
rhornung67, line 41: Would it be good to make a test or example code for this for each back-end?

rhornung67 (dismissed):

include/camp/resource/sycl.hpp
```diff
@@ -34,16 +34,28 @@ namespace resources
...
       SyclEvent(Sycl& res);
 
-      bool check() const { return true; }
+      bool check() const
+      {
+        return m_event.get_info<sycl::info::event::command_execution_status>()
+            == sycl::info::event_command_status::complete;
+      }
+
+      void wait() const { m_event.wait(); }
```
MrBurmark: It appears that `sycl::event::wait` is a non-const member function, so this doesn't compile here. However it did compile on corona for some reason.

MrBurmark: I made a copy to get a non-const event and work around this restriction.

rhornung67: Hmm. What SYCL header file are you looking at? We are using an older version of the oneapi complier on corona.

include/camp/resource/sycl.hpp
MrBurmark: Indeed it looks like we don't actually have tests for Event or at least we don't have tests for all of event. I'll work on adding those.

adayton1 (dismissed):

include/camp/resource/sycl.hpp
rhornung67: Thank you @MrBurmark

include/camp/resource/sycl.hpp
MrBurmark: I think it probably just didn't notice the issue since the function was inline and never called.

adayton1 (dismissed):

MrBurmark:
@homerdin Does this seem like a reasonable way to implement events on top of sycl?

homerdin:
@MrBurmark Yes, I think this is a good way to go.

include/camp/resource/sycl.hpp
```diff
@@ -34,16 +34,32 @@ namespace resources
...
       SyclEvent(Sycl& res);
 
-      bool check() const { return true; }
+      bool check() const
+      {
+        return m_event.get_info<sycl::info::event::command_execution_status>()
+            == sycl::info::event_command_status::complete;
+      }
+
+      void wait() const { sycl::event(m_event).wait(); }
```
adayton1: Isn't m_event already a sycl::event?

MrBurmark: I should have added a comment, this is here because `event.wait()` is a non-const member function.

MrBurmark:
```suggestion
      void wait() const
      {
        // create a copy of the event as wait is a non-const member function.
        sycl::event(m_event).wait();
      }
```

MrBurmark: I suppose that I could also make `m_event` mutable or use a const_cast here. Not quite sure what would be best, though avoiding incrementing and decrementing the reference count of the memory referred to by m_event would be nice.

adayton1: Maybe it should be mutable, then.

MrBurmark: Yep, I'll go to that then.

MrBurmark: done

adayton1 (approved):

rhornung67 (approved):

PR #204, opened 2026-05-28
Event in Containers

MrBurmark:
Make events usable as keys in some containers like unordered_map.
Add hashability and comparability.

I also made Event and Resource safe to use when empty, but I might break that off into a separate PR as its orthogonal to these changes.

### DIFF
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -64,6 +64,8 @@ namespace resources
 
       CudaEvent(Cuda &res);
 
+      Platform get_platform() const { return Platform::cuda; }
+
       bool check() const
       {
         return (CAMP_CUDA_API_INVOKE_AND_CHECK_RETURN(cudaEventQuery, m_event)
@@ -76,6 +78,20 @@ namespace resources
       }
 
       cudaEvent_t getCudaEvent_t() const { return m_event; }
+
+      /*
+       * \brief Compares two events to see if they are equal
+       *
+       * \return True or false depending on if this is the same event
+       */
+      friend inline bool operator==(CudaEvent const& lhs, CudaEvent const& rhs) = default;
+
+      size_t get_hash() const
+      {
+        const size_t platform_type = size_t(get_platform()) << 32;
+        size_t hash = std::hash<cudaEvent_t>{}(m_event);
+        return platform_type | (hash & 0xFFFFFFFF);
+      }
 
     private:
       cudaEvent_t m_event;
@@ -315,6 +331,26 @@ namespace resources
 }  // namespace resources
 }  // namespace camp
 
+namespace std
+{
+
+/*
+ * \brief Specialization of std::hash for camp::resources::CudaEvent
+ *
+ * Provides a hash function for cuda typed event objects, enabling their use
+ * as keys in unordered associative containers (std::unordered_map,
+ * std::unordered_set, etc.)
+ *
+ * \return A size_t hash value
+ */
+template <>
+struct hash<camp::resources::CudaEvent> {
+  std::size_t operator()(const camp::resources::CudaEvent &e) const
+  {
+    return e.get_hash();
+  }
+};
+
 /*
  * \brief Specialization of std::hash for camp::resources::Cuda
  *
@@ -324,8 +360,6 @@ namespace resources
  *
  * \return A size_t hash value
  */
-namespace std
-{
 template <>
 struct hash<camp::resources::Cuda> {
   std::size_t operator()(const camp::resources::Cuda &c) const
@@ -333,6 +367,7 @@ struct hash<camp::resources::Cuda> {
     return c.get_hash();
   }
 };
+
 }  // namespace std
 
 #endif  // #ifdef CAMP_ENABLE_CUDA
--- a/include/camp/resource/cuda.hpp
+++ b/include/camp/resource/cuda.hpp
@@ -286,17 +286,7 @@ namespace resources
        *
        * \return True or false depending on if it is the same stream
        */
-      bool operator==(Cuda const &c) const
-      {
-        return (get_stream() == c.get_stream());
-      }
-
-      /*
-       * \brief Compares two (Cuda) resources to see if they are NOT equal
-       *
-       * \return Negation of == operator
-       */
-      bool operator!=(Cuda const &c) const { return !(*this == c); }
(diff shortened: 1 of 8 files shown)


include/camp/resource/cuda.hpp
```diff
@@ -77,6 +79,20 @@ namespace resources
...
+       * \brief Compares two events to see if they are equal
+       *
+       * \return True or false depending on if this is the same event
+       */
+      friend inline bool operator==(CudaEvent const& lhs, CudaEvent const& rhs) = default;
+
+      size_t get_hash() const
+      {
+        const size_t platform_type = size_t(get_platform()) << 32;
+        size_t hash = std::hash<cudaEvent_t>{}(m_event);
```
adayton1, line 93: Is cudaEvent_t a pointer?

MrBurmark: Yes, its a typedef to a pointer.

include/camp/resource/cuda.hpp
```diff
@@ -334,15 +360,14 @@ namespace resources
...
-{
 template <>
 struct hash<camp::resources::Cuda> {
   std::size_t operator()(const camp::resources::Cuda &c) const
   {
     return c.get_hash();
   }
 };
+
 }  // namespace std
```
kab163, line 373: Are there extraneous `}`s at the end of these files?

kab163: Oh, nevermind, it's just in how the diff's are displayed!

include/camp/resource/cuda.hpp
rhornung67: Would it be better to rename/alias it to something like `ptr_CudaEvent_t` for clarity?

include/camp/resource/event.hpp
```diff
@@ -41,41 +45,136 @@ namespace resources
...
         auto result = dynamic_cast<EventModel<T> *>(m_value.get());
+        if (!result) {
+          return nullptr;
+        }
         return result->get();
       }
 
       template <typename T>
-      T get()
+      T const*try_get() const
```
rhornung67, line 71:
```suggestion
      T const* try_get() const
```

include/camp/resource/cuda.hpp
```diff
@@ -77,6 +79,20 @@ namespace resources
 
       cudaEvent_t getCudaEvent_t() const { return m_event; }
 
+      /*
+       * \brief Compares two events to see if they are equal
```
rhornung67: This definition of equality seems inconsistent with the comment here:

https://github.com/llnl/camp/pull/204/changes#diff-ada08ef8bc61b00c4422ad9adfdccc9cf3b029c312e7412fb714fb1ce2a8faa7R133

include/camp/resource/event.hpp
adayton1, line 71: Could you add whitespace at least on one side of the asterisk? I think there are a few other places like this that could also use whitespace.

include/camp/resource/event.hpp
```diff
@@ -41,41 +45,136 @@ namespace resources
...
+        return m_value ? m_value->get_platform() : Platform::undefined;
+      }
+
+      bool check() const { return m_value ? m_value->check() : true; }
+
+      void wait() const { if (m_value) { m_value->wait(); } }
+
+      /*
+       * \brief Compares two Events to see if they are equal. Two Events
+       * are equal if they have the platform and same event
```
rhornung67: I'm confused by the statement "have the same platform and same event". What does it mean for an event to have an event? Should that be changed to something like "operation"?

include/camp/resource/host.hpp
```diff
@@ -28,9 +28,24 @@ namespace resources
...
       HostEvent() {}
 
+      Platform get_platform() const { return Platform::host; }
+
       bool check() const { return true; }
 
       void wait() const {}
+
+      /*
+       * \brief Compares two events to see if they are equal
```
rhornung67: Similar to my comment above. To me, it appears that we are using two different definitions of equality, one being that the events are the same objects and the other being that they have the same platform and refer to the same operation.  

This sort of thing occurs in multiple places.

rhornung67 (changes requested):
I think we need to make it more clear what we mean when we say two events are equal; when `e1 == e2` is true.

include/camp/resource/event.hpp
rhornung67: @adayton1 thanks for catching that. I suggested the same change above. 😄

include/camp/resource/cuda.hpp
MrBurmark: I don't think that is more clear, I could add a comment though.

include/camp/resource/event.hpp
MrBurmark: This is the type erased event class that holds a typed event.

include/camp/resource/cuda.hpp
adayton1: I think a typedef would be worse. But a comment saying that the hash is based on the pointer address might help.

include/camp/resource/cuda.hpp
```diff
@@ -77,6 +79,20 @@ namespace resources
...
+       *
+       * \return True or false depending on if this is the same event
+       */
+      friend inline bool operator==(CudaEvent const& lhs, CudaEvent const& rhs) = default;
+
+      size_t get_hash() const
+      {
+        const size_t platform_type = size_t(get_platform()) << 32;
```
adayton1, line 92: Is the platform really that helpful in the hash? If you have two different kinds of events with the same pointer address at the same time, wouldn't that mean you've got a bug in your program?

include/camp/resource/cuda.hpp
MrBurmark: Added a comment about the type where the member is declared.

include/camp/resource/cuda.hpp
MrBurmark: Updated the equality comments.

include/camp/resource/cuda.hpp
MrBurmark: I'm following what was done for the resource hash.

include/camp/resource/event.hpp
MrBurmark: Updated.

adayton1 (approved):

rhornung67 (approved):

PR #205, opened 2026-05-29
Add cmake option for CAMP_USE_PLATFORM_DEFAULT_STREAM

helloworld922:


CMakeLists.txt
```diff
@@ -119,6 +119,13 @@ endif ()
 
 # end backends
 
+option(CAMP_USE_PLATFORM_DEFAULT_STREAM "Whether CAMP should use the default device streams (On) or create separate streams (Off)" Off)
+
+set(CAMP_USE_PLATFORM_DEFAULT_STREAM_NUM 0)
```
white238: Nitpick: should set this to `OFF` and `ON`. If you use `cmakedefine01`, CMake will also set it to `0` and `1` respectively w/o having the additional `@CAMP_USE_PLATFORM_DEFAULT_STREAM_NUM@`. If that is really needed as opposed to just having it defined in preprocessor.

https://cmake.org/cmake/help/latest/command/configure_file.html#transformations

artv3:
Used internally: https://github.com/llnl/camp/blob/56b194f3ea3d0e9d6e2f236de39a4886063e4968/include/camp/resource/hip.hpp#L187C5-L187C37 

@helloworld922 , is the issue that we do not have a CMAKE variable to define it?

helloworld922:
Yes, the define is supposed to be user defined if they want to use default streams, but unless you properly forward the compiler define to all downstream dependencies you either end up with an ODR violation and/or not having the correct behavior of which stream kernels are being run on. This just makes sure that the define is baked into the installed RAJA headers so everyone picks up a consistent definition.

include/camp/config.in.hpp
```diff
@@ -15,6 +15,7 @@
 #cmakedefine CAMP_ENABLE_HIP
 #cmakedefine CAMP_ENABLE_SYCL
 #cmakedefine CAMP_WIN_STATIC_BUILD
+#cmakedefine01 CAMP_USE_PLATFORM_DEFAULT_STREAM
```
trws: I'd rather this be wrapped in an `ifndef CAMP_USE_PLATFORM_DEFAULT_STREAM` so we don't have a potential behavior change for projects that set this currently.

helloworld922: I moved the check originally in `defines.hpp` to `config.in.hpp`

MrBurmark (dismissed):

adayton1 (dismissed):

include/camp/config.in.hpp
```diff
@@ -15,6 +15,9 @@
 #cmakedefine CAMP_ENABLE_HIP
 #cmakedefine CAMP_ENABLE_SYCL
 #cmakedefine CAMP_WIN_STATIC_BUILD
+#ifndef CAMP_USE_PLATFORM_DEFAULT_STREAM
+#cmakedefine01 CAMP_USE_PLATFORM_DEFAULT_STREAM
```
MrBurmark: Do we want to add a warning or at least a comment that if CAMP_USE_PLATFORM_DEFAULT_STREAM is inconsistently set in different translation units that it will cause an ODR violation?

rhornung67: I think a warning would be helpful.

helloworld922: added a warning path check if the user defines something different

rhornung67 (dismissed):

include/camp/config.in.hpp
```diff
@@ -15,6 +15,15 @@
...
+#else
+#cmakedefine01 CAMP_USE_PLATFORM_DEFAULT_STREAM_INTERNAL_CHECK
```
adayton1: I don't get this change. CAMP_USE_PLATFORM_DEFAULT_STREAM_INTERNAL_CHECK is not a CMake variable.

include/camp/config.in.hpp
adayton1: Perhaps it would be better to just error if CAMP_USE_PLATFORM_DEFAULT_STREAM is already defined.

include/camp/config.in.hpp
helloworld922: Oops, I forgot to also add the CMake change. This should generate the correct config file now and give the appropriate mismatch warning.

include/camp/config.in.hpp
helloworld922: I'm fine with either a warning or an error, which one do @MrBurmark and @rhornung67 prefer?

MrBurmark: A warning is probably good.

MrBurmark (dismissed):

adayton1 (dismissed):

rhornung67 (dismissed):

include/camp/config.in.hpp
```diff
@@ -15,6 +15,15 @@
...
+#if CAMP_USE_PLATFORM_DEFAULT_STREAM != CAMP_USE_PLATFORM_DEFAULT_STREAM_INTERNAL_CHECK
+#warning "CAMP_USE_PLATFORM_DEFAULT_STREAM mismatch, potential ODR violation"
```
rhornung67: Make this an error

MrBurmark: @helloworld922 If you no longer will need to define CAMP_USE_PLATFORM_DEFAULT_STREAM from outside of this header we were thinking about disallowing that to avoid ODR violations.

helloworld922: I think that would be ideal. I moved the check to a part of `config.in.hpp` which should only be defined once. It still supports users having manually defined `CAMP_USE_PLATFORM_DEFAULT_STREAM` for backwards compatibility, but it must match how RAJA is configured to be built.

Alternatively, I could just remove this and force all users to not have `CAMP_USE_PLATFORM_DEFAULT_STREAM` externally defined at all.

adayton1: I like the idea of a hard error if users have CAMP_USE_PLATFORM_DEFAULT_STREAM externally defined:

```
#if defined(CAMP_USE_PLATFORM_DEFAULT_STREAM)
#error "Manually defining CAMP_USE_PLATFORM_DEFAULT_STREAM is not allowed because of the potential for ODR violations. CAMP now supports a CMake configuration option that should be used instead."
#endif

#cmakedefine01 CAMP_USE_PLATFORM_DEFAULT_STREAM
```

include/camp/config.in.hpp
```diff
@@ -34,3 +34,16 @@
 #else
 #define CAMP_DLL_EXPORT
 #endif
+
+#ifndef CAMP_CONFIG_HPP
```
adayton1: @trws, @MrBurmark should this header guard be at the beginning of the file or after the CAMP_CONFIG_OVERRIDE section?

MrBurmark: Previously there was no header guard at all, so it would make sense to me to add one around everything. @trws was there a specific reason not to guard this header?

MrBurmark: I assume that even if you are using the override you can only want to config camp once per translation unit.

include/camp/config.in.hpp
adayton1, line 12:
```suggestion
#ifndef CAMP_CONFIG_HPP
#define CAMP_CONFIG_HPP

```

include/camp/config.in.hpp
```diff
@@ -34,3 +34,16 @@
...
+#define CAMP_CONFIG_HPP
+
+// stuff that should always be only defined once
+
```
adayton1:
```suggestion
```

include/camp/config.in.hpp
```diff
@@ -6,6 +6,8 @@
 //
 // SPDX-License-Identifier: (BSD-3-Clause)
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
+#ifndef CAMP_CONFIG_HPP
```
adayton1, line 10:
```suggestion

#ifndef CAMP_CONFIG_HPP
```

include/camp/config.in.hpp
```diff
@@ -34,3 +37,11 @@
 #else
 #define CAMP_DLL_EXPORT
 #endif
+
+#ifdef CAMP_USE_PLATFORM_DEFAULT_STREAM
+#error "Manually defining CAMP_USE_PLATFORM_DEFAULT_STREAM is not allowed because of the potential for ODR violations. CAMP now supports a CMake configuration option that should be used instead."
```
adayton1:
```suggestion
#error "Manually defining CAMP_USE_PLATFORM_DEFAULT_STREAM is not allowed because of the potential for ODR violations. CAMP supports a CMake configuration option that should be used instead."
```

include/camp/config.in.hpp
```diff
@@ -34,3 +37,11 @@
 #else
 #define CAMP_DLL_EXPORT
 #endif
+
+#ifdef CAMP_USE_PLATFORM_DEFAULT_STREAM
```
adayton1:
```suggestion
#if defined(CAMP_USE_PLATFORM_DEFAULT_STREAM)
```

adayton1 (approved):

MrBurmark (approved):

PR #206, opened 2026-06-03
Move Only Events

MrBurmark:
Make typed events move only so they can own the native events that they contain and destroy those native events properly. Erased events (Event) are not move only and still have reference semantics as they are implemented with a shared_ptr.

This also other things I found along the way. For example some functions only took the erased version of a resource or event so I added an overload for the typed version for performance. I also updated the organization of some of the type traits and concepts used by the resources/events. I also updated the tests to test more things.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -36,6 +36,7 @@ namespace resources
     public:
       using event_type = Event;
 
+      Resource() = default;
       Resource(Resource &&) = default;
       Resource(Resource const &) = default;
       Resource &operator=(Resource &&) = default;
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -209,6 +209,8 @@ namespace resources
         m_value->wait();
       }
 
+      explicit operator bool() const { return static_cast<bool>(m_value); }
+
       /*
        * \brief Compares two Resources to see if they are equal. Two Resources
        * are equal if they have the platform and same stream/queue
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -41,11 +41,7 @@ namespace resources
       Resource &operator=(Resource &&) = default;
       Resource &operator=(Resource const &) = default;
 
-      template <typename T,
-                typename = typename std::enable_if_t<
-                    !std::is_same_v<typename std::decay_t<T>, Resource>
-                    && is_concrete_resource_v<T>>
-                >
+      template <camp::concepts::ConcreteResource T>
       Resource(T &&value)
       {
         m_value.reset(new ContextModel<type::ref::rem<T>>(forward<T>(value)));
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -276,7 +276,7 @@ namespace resources
       class ContextModel final : public ContextInterface
       {
       public:
-        ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
+        explicit ContextModel(T const &modelVal) : m_modelVal(modelVal) {}
 
         Platform get_platform() const override
         {
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -181,7 +181,7 @@ namespace resources
         return m_value->get_event_erased();
       }
 
-      void wait_for(Event& e)
+      void wait_for(Event const& e)
       {
         if (!m_value) {
             e.wait();
@@ -280,7 +280,7 @@ namespace resources
 
         virtual Event get_event() = 0;
         virtual Event get_event_erased() = 0;
-        virtual void wait_for(Event& e) = 0;
+        virtual void wait_for(Event const& e) = 0;
         virtual void wait() = 0;
       };
 
@@ -337,7 +337,7 @@ namespace resources
           return m_modelVal.get_event_erased();
         }
 
-        void wait_for(Event& e) override { m_modelVal.wait_for(e); }
+        void wait_for(Event const& e) override { m_modelVal.wait_for(e); }
 
         void wait() override { m_modelVal.wait(); }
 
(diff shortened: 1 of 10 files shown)


include/camp/resource/cuda.hpp
```diff
@@ -193,25 +241,41 @@ namespace resources
...
-      Event get_event_erased() { return Event{CudaEvent(*this)}; }
+      Event get_event_erased() { return Event{get_event()}; }
 
       void wait()
       {
         auto d{device_guard(device)};
         CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamSynchronize, stream);
       }
 
+      void wait_for(std::nullptr_t) {}
```
adayton1: Do we really need this overload? Or should it be explicitly disallowed at compile time?

include/camp/resource/host.hpp
```diff
@@ -65,15 +94,27 @@ namespace resources
...
+        }
+        e->wait();
+      }
+
+      void wait_for(Event *e)
+      {
+        if (!e) {
+          return;
+        }
+        e->wait();
```
adayton1: No cast to the concrete type?

include/camp/resource/event.hpp
```diff
@@ -152,6 +138,20 @@ namespace resources
         return false;
       }
 
+      template < camp::concepts::ConcreteEvent Evt >
+      friend inline bool operator==(Event const &lhs, Evt const &rhs)
```
adayton1, line 142: What about the case where the concrete event is on the left hand side?

include/camp/resource.hpp
```diff
@@ -266,6 +223,20 @@ namespace resources
         return false;
       }
 
+      template < camp::concepts::ConcreteResource Res >
+      friend inline bool operator==(Resource const &lhs, Res const &rhs)
```
adayton1, line 226: What about where the order is reversed?

test/resource.cpp
```diff
@@ -194,6 +201,145 @@ TEST(CampEvent, GetPlatform)
...
+  vec.emplace_back(h);
+  vec.emplace_back(d1);
+  vec.emplace_back(r1);
+
+  // Typed
+  rvec.emplace_back(Res());
+  rvec.emplace_back(d2.get<Res>());
+  rvec.emplace_back(r2);
+
+  // Verify using Resource in a vector to works
```
adayton1: This comment doesn't make sense to me.

include/camp/resource/cuda.hpp
MrBurmark: This is here because we now have the overloads for the concrete and typed event so passing in an explicit nullptr would be ambiguous. This is one way to fix this ambiguity, explicitly deleting this would be another. I have always thought it was weird that we passed in a pointer instead of a reference here to begin with, so we could also do that instead if we're ok with more breaking changes.

include/camp/resource/host.hpp
MrBurmark: We could defer to the concrete type here as in the other implementations, I just know that here we are calling wait either way.

include/camp/resource/event.hpp
MrBurmark: I didn't do that because c++20 the compiler can reverse the order of the arguments. However, since you mentioned this I'm not sure that it will do that in this case because the (erased,erased) version works for (typed,erased). I will check what it does and add the other version if necessary.

test/resource.cpp
MrBurmark:
```suggestion
  // Verify using Resource in a vector works
```

include/camp/resource/host.hpp
adayton1: Yeah I was more confused that it didn't match the other implementations.

include/camp/resource/cuda.hpp
adayton1: I think passing in a reference would be preferable. Or explicitly deleting it. I don't think there is any reason for someone to call this with the literal text `nullptr`.

include/camp/resource/host.hpp
MrBurmark: I'll make it consistent.

include/camp/resource/event.hpp
MrBurmark: It does use the more specific one even if it has to reverse the argument order.

include/camp/resource/cuda.hpp
MrBurmark: I'll make new overloads for references. Then I'm not sure if its better to remove the pointer version or just deprecate it.

rhornung67: Do you know how many projects are using the `wait_for()` method? If it's not used much or at all, then it makes sense to remove the pointer version(s).

MrBurmark: I'm not aware of anyone using it, so I went ahead and removed the pointer version.

include/camp/resource/cuda.hpp
```diff
@@ -16,8 +16,10 @@
 
 #include <cuda_runtime.h>
 
+#include <cstddef>
```
adayton1, line 19: What does cstddef bring that you need?

adayton1: You might be able to remove it in several headers.

adayton1 (approved):

include/camp/resource/cuda.hpp
MrBurmark: We use size_t in many of these so we probably still should have it. We should probably even use std::size_t.

rhornung67: I think it's best to use fully-qualified names in header files to avoid `using` directives in those.

rhornung67 (approved):

PR #207, opened 2026-07-06
Resource allocator

bechols97:
Moves the resource allocator implementation from https://github.com/llnl/RAJA/pull/1832 to camp. The goal of this is to support a C++ `std::allocator`-like allocator for resources.

### DIFF
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -24,7 +24,6 @@
 
 // last to ensure we don't hide breakage in the others
 #include "camp/resource/platform.hpp"
-#include "camp/resource/resource_allocator.hpp"
 
 namespace camp
 {
@@ -423,5 +422,7 @@ struct hash<camp::resources::Resource> {
 #if defined(CAMP_HAVE_OMP_OFFLOAD)
 #include "camp/resource/omp_target.hpp"
 #endif
+
+#include "camp/resource/resource_allocator.hpp"
 
 #endif /* __CAMP_RESOURCE_HPP */
--- a/include/camp/resource.hpp
+++ b/include/camp/resource.hpp
@@ -19,10 +19,12 @@
 #include "camp/config.hpp"
 #include "camp/defines.hpp"
 #include "camp/helpers.hpp"
+
 #include "camp/resource/event.hpp"
 
 // last to ensure we don't hide breakage in the others
 #include "camp/resource/platform.hpp"
+#include "camp/resource/resource_allocator.hpp"
 
 namespace camp
 {
@@ -48,7 +50,7 @@ namespace resources
       }
 
       template <typename T>
-      T *try_get()
+      T* try_get()
       {
         if (!m_value) {
           return nullptr;
@@ -61,7 +63,7 @@ namespace resources
       }
 
       template <typename T>
-      T const *try_get() const
+      T const* try_get() const
       {
         if (!m_value) {
           return nullptr;
@@ -74,9 +76,9 @@ namespace resources
       }
 
       template <typename T>
-      T &get() &
+      T& get() &
       {
-        T *result = try_get<T>();
+        T* result = try_get<T>();
         if (result == nullptr) {
           ::camp::throw_re("Incompatible Resource type get cast.");
         }
@@ -84,9 +86,9 @@ namespace resources
       }
 
       template <typename T>
-      T const &get() const &
+      T const& get() const&
       {
-        T const *result = try_get<T>();
+        T const* result = try_get<T>();
         if (result == nullptr) {
           ::camp::throw_re("Incompatible Resource type get cast.");
         }
@@ -96,7 +98,7 @@ namespace resources
       template <typename T>
       T get() &&
       {
-        T *result = try_get<T>();
+        T* result = try_get<T>();
         if (result == nullptr) {
           ::camp::throw_re("Incompatible Resource type get cast.");
         }
@@ -180,10 +182,10 @@ namespace resources
         return m_value->get_event_erased();
       }
 
-      void wait_for(Event const &e)
+      void wait_for(Event const& e)
       {
         if (!m_value) {
-          e.wait();
+            e.wait();
           return;
         }
         m_value->wait_for(e);
@@ -211,7 +213,8 @@ namespace resources
         if (!lhs.m_value && !rhs.m_value) {
           return true;
         }
-        if ((!lhs.m_value && rhs.m_value) || (lhs.m_value && !rhs.m_value)) {
+        if ((!lhs.m_value && rhs.m_value) ||
+            (lhs.m_value && !rhs.m_value)) {
           return false;
         }
         if (lhs.get_platform() == rhs.get_platform()) {
@@ -220,7 +223,7 @@ namespace resources
         return false;
       }
 
-      template <camp::concepts::ConcreteResource Res>
+      template < camp::concepts::ConcreteResource Res >
       friend inline bool operator==(Resource const &lhs, Res const &rhs)
       {
         if (!lhs.m_value) {
@@ -269,7 +272,7 @@ namespace resources
 
         virtual Event get_event() = 0;
         virtual Event get_event_erased() = 0;
-        virtual void wait_for(Event const &e) = 0;
+        virtual void wait_for(Event const& e) = 0;
         virtual void wait() = 0;
       };
 
@@ -326,13 +329,13 @@ namespace resources
           return m_modelVal.get_event_erased();
         }
 
-        void wait_for(Event const &e) override { m_modelVal.wait_for(e); }
+        void wait_for(Event const& e) override { m_modelVal.wait_for(e); }
 
         void wait() override { m_modelVal.wait(); }
 
-        const T *get() const { return &m_modelVal; }
+        const T* get() const { return &m_modelVal; }
 
-        T *get() { return &m_modelVal; }
+        T* get() { return &m_modelVal; }
 
       private:
         T m_modelVal;
@@ -342,7 +345,8 @@ namespace resources
     };
 
     template <typename Res>
-    struct EventProxy : ::camp::resources::detail::EventProxyBase {
+    struct EventProxy : ::camp::resources::detail::EventProxyBase
+    {
       using native_event = typename Res::event_type;
 
       EventProxy(EventProxy &&) = default;
@@ -353,24 +357,27 @@ namespace resources
       EventProxy(Res r) : resource_{move(r)} {}
 
       native_event get()
-        requires(!std::same_as<native_event, Event>)
+      requires (!std::same_as<native_event, Event>)
       {
         return resource_.get_event();
       }
 
       Event get()
-        requires(std::same_as<native_event, Event>)
+      requires (std::same_as<native_event, Event>)
       {
         return resource_.get_event_erased();
       }
 
       operator native_event()
-        requires(!std::same_as<native_event, Event>)
+      requires (!std::same_as<native_event, Event>)
       {
         return resource_.get_event();
       }
 
-      operator Event() { return resource_.get_event_erased(); }
+      operator Event()
+      {
+        return resource_.get_event_erased();
+      }
 
       Res resource_;
     };
@@ -416,7 +423,5 @@ struct hash<camp::resources::Resource> {
 #if defined(CAMP_HAVE_OMP_OFFLOAD)
 #include "camp/resource/omp_target.hpp"
 #endif
-
-#include "camp/resource/resource_allocator.hpp"
 
 #endif /* __CAMP_RESOURCE_HPP */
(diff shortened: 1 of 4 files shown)


include/camp/resource/resource_allocator.hpp
```diff
@@ -0,0 +1,108 @@
...
+    void deallocate(value_type* ptr, std::size_t) noexcept
+    {
+      m_res.deallocate(ptr, m_mem_type);
+    }
+
+    [[nodiscard]]
+    Resource const& get_resource() const { return m_res; }
+
+    [[nodiscard]]
+    Resource get_resource() { return m_res; }
```
MrBurmark: Should this return a reference?

include/camp/resource/resource_allocator.hpp
```diff
@@ -0,0 +1,108 @@
...
+
+    [[nodiscard]]
+    camp::resources::MemoryAccess get_mem_access() const { return m_mem_type; }
+
+    [[nodiscard]]
+    camp::resources::MemoryAccess get_mem_access() { return m_mem_type; }
```
MrBurmark: Is the const version all we need?

test/resource_allocator.cpp
```diff
@@ -0,0 +1,306 @@
...
+#ifdef CAMP_HAVE_OMP_OFFLOAD
+  test_move_assignment<Omp>();
+#endif
+#ifdef CAMP_HAVE_SYCL
+  test_move_assignment<Sycl>();
+#endif
+}
+
+template <typename Res>
+void test_get_resource(Res res)
```
MrBurmark, line 174: Do we want to test the const and non-const methods?

include/camp/resource/resource_allocator.hpp
MrBurmark: I'm mainly thinking about using the non-const member functions like wait, get_event, allocate, etc. on a resource stored in an allocator.

bechols97: I didn't return a reference here due to the users being able to modify the internal data to this allocator, which I was a bit hesitant on allowing. But, if it is necessary for wait, get_event, etc., then I can change it.

MrBurmark: Ya, it does open up that possibility. I see that `std::vector::get_allocator` returns by value, so that does appear to be the way this sort of function is implemented. These should be cheap to copy anyway.

MrBurmark (dismissed):

rhornung67 (dismissed):

test/resource_allocator.cpp
```diff
@@ -0,0 +1,285 @@
...
+
+  int* ptr = alloc1.allocate(num);
+
+  EXPECT_TRUE(ptr != nullptr);
+
+  alloc1.deallocate(ptr, num);
+}
+
+//
+TEST(CampResourceAllocator, Allocate)
```
MrBurmark, line 384: How about adding a test that the allocator with the host resource works with some std containers?

bechols97: Added tests for `std::vector` using the `ResourceAllocator`. If there were any particular containers that you would like me to add, then let me know.

MrBurmark: How about adding making another vector by doing things like copy, move, assignment, move assignment to exercise that part of the allocator infrastructure?

test/resource_allocator.cpp
```diff
@@ -0,0 +1,320 @@
...
+
+  for (std::size_t i = 0; i < vec.size(); ++i) {
+    EXPECT_TRUE(vec[i] == static_cast<int>(i));
+  }
+}
+
+//
+TEST(CampResourceAllocator, Vector)
+{
+  test_allocate<Host>();
```
MrBurmark, line 458: How about adding a test for a generic Resource as well, perhaps in all of the tests.

include/camp/resource/resource_allocator.hpp
```diff
@@ -0,0 +1,109 @@
...
+struct ResourceAllocator
+{
+  template<typename T>
+  struct allocator
+  {
+    using value_type = T;
+
+    allocator()
+        : allocator(Resource::get_default(),
+                    camp::resources::MemoryAccess::Pinned)
```
adayton1: Instead of hardcoding this to pinned, shouldn't this be based on some resource default?

test/resource_allocator.cpp
```diff
@@ -0,0 +1,320 @@
...
+#endif
+#ifdef CAMP_HAVE_SYCL
+  test_allocate<Sycl>();
+#endif
+}
+
+template <typename Res>
+void test_vector()
+{
+  static constexpr std::size_t num = 5;
```
artv3, line 404: is static needed here?

test/resource_allocator.cpp
```diff
@@ -0,0 +1,320 @@
...
+#endif
+#ifdef CAMP_HAVE_SYCL
+  test_rebind<Sycl>();
+#endif
+}
+
+template <typename Res>
+void test_allocate()
+{
+  static constexpr std::size_t num = 5;
```
artv3, line 366: constexpr should be enough right?

MrBurmark: Using static with constexpr local variables helps to tell the compiler that the value will live indefinitely and helps to convince the compiler to actually treat them as compile time constants even if they are not used as compile time constants.

include/camp/resource/resource_allocator.hpp
bechols97: Do the resources have some getter for a default for the `MemoryAccess`? If not, I can use `Device` instead since that is the default in the generic resource allocate.

MrBurmark: I don't think there is a global constant defining what is the default if that's what you mean.

MrBurmark: Given that this is camp, I agree its best to use the same default as the resources. You could even make a default in the enum if you wanted to centralize that decision.

test/resource_allocator.cpp
bechols97: While it probably isn't needed for the test, the `static` here can help the compiler optimize as Jason mentioned.

test/resource_allocator.cpp
```diff
@@ -0,0 +1,449 @@
...
+void test_compare()
+{
+  // Specific resource
+  camp::ResourceAllocator<int, Res> alloc1{Res(), MemoryAccess::Pinned};
+  camp::ResourceAllocator<int, Res> alloc2{Res(), MemoryAccess::Device};
+  camp::ResourceAllocator<int, Res> alloc3{Res(), MemoryAccess::Managed};
+
+  ASSERT_EQ(alloc1, alloc1);
+  ASSERT_NE(alloc1, alloc2);
+  ASSERT_NE(alloc1, alloc3);
```
adayton1, line 278: Would it also make sense to check that alloc2 == alloc2, alloc2 != alloc3, and alloc3 == alloc3?

adayton1: Similarly for the generic resource checks below.

adayton1 (dismissed):

test/resource_allocator.cpp
bechols97: Oh good catch, looks like I forgot those

include/camp/resource/resource_allocator.hpp
bechols97: Updated the default to be the same as the `Resource.allocate` function.

artv3 (dismissed):

adayton1 (dismissed):

rhornung67 (dismissed):
LGTM. Thank you @bechols97

include/camp/resource/resource_allocator.hpp
```diff
@@ -0,0 +1,104 @@
...
+{
+template<typename Resource>
+struct ResourceAllocator
+{
+  template<typename T>
+  struct allocator
+  {
+    using value_type = T;
+
+    allocator(Resource res = Resource {},
```
MrBurmark: Does this have to be default construct-able? For the erased resource `resources::Resource` this will generate a resource pointing to nothing and will throw instead of being able to allocate/free.

test/resource_allocator.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+template <typename Res>
+void test_get_mem_access()
+{
+  // Specific resource
+  camp::ResourceAllocator<int, Res> alloc1{Res(), MemoryAccess::Pinned};
+  camp::ResourceAllocator<int, Res> alloc2{Res(), MemoryAccess::Device};
+  camp::ResourceAllocator<int, Res> alloc3{Res(), MemoryAccess::Managed};
+
+  ASSERT_EQ(alloc1.get_mem_access(), alloc1.get_mem_access());
```
MrBurmark, line 222: Should these test for the specific MemoryAccess they were constructed with?

test/resource_allocator.cpp
```diff
@@ -0,0 +1,459 @@
...
+  // Specific resource
+  camp::ResourceAllocator<int, Res> alloc1{Res()};
+  camp::ResourceAllocator<double, Res> alloc2{alloc1};
+
+  ASSERT_EQ(alloc1, alloc2);
+  ASSERT_EQ(alloc1.get_resource(), alloc2.get_resource());
+  ASSERT_EQ(alloc1.get_mem_access(), alloc2.get_mem_access());
+
+  // Generic resource
+  camp::ResourceAllocator<int, Resource> alloc3{Res()};
```
MrBurmark, line 330: Is there a test to construct a `camp::ResourceAllocator<int, Resource>` from a `camp::ResourceAllocator<int, Res>`?

include/camp/resource/resource_allocator.hpp
trws: It doesn't in a "will this comply with the spec" sense. Umpire's base allocator and all stateless allocators have one though, so it might be worth taking a quick look around our codes to see if a requirement for it has crept in somewhere

bechols97: It is worth noting that if the default construct is not there. Then, something like the example below won't compile.
```
std::vector<int, camp::ResourceAllocator<int, camp::resources::Resource>> vec;
...
vec = std::vector<int, camp::ResourceAllocator<int, camp::resources::Resource>>(
  num_elements, allocator);
````

Though as @trws mentioned, a default construct is not required by the spec.

Comment (trws) [reply include/camp/resource/resource_allocator.hpp]: Yeah, that's a great example of the issue.  It's really easy to have a pattern like that show up in code rather than:

```
auto vec = std::vector<int, camp::ResourceAllocator<int, camp::resources::Resource>>(
  num_elements, allocator);
```

Which wouldn't have the issue.  So many allocators have a default constructor that I'd be kinda surprised if it didn't come up TBH.  Perhaps it would be worth having the default constructor create something that dies with a clear and explicit error message if anything attempts to use it without properly initializing it?

Comment (MrBurmark) [reply include/camp/resource/resource_allocator.hpp]: An exception is thrown if you try to call allocate on an empty Resource.
https://github.com/llnl/camp/blob/788aec085bba59548fd778517c10a46279bc8cf2/include/camp/resource.hpp#L119

Comment (bechols97) [reply include/camp/resource/resource_allocator.hpp]: Normally, I tend to prefer compiler errors when possible. However, in this case, my preference would be to have a default construct that throws an exception if not properly initialized. The pattern that was in the example tends to show up more often than it really should.

Comment (trws) [reply include/camp/resource/resource_allocator.hpp]: Fair enough. With the removals in 20 that actually covers it all, I was thinking even in cases like rebind, but those are no longer a thing I guess so yeah.  Does that work for you @MrBurmark?

Comment (MrBurmark) [reply include/camp/resource/resource_allocator.hpp]: Sure its a bit concerning but at least it will throw instead of trying to dereference a null ptr.

File: test/resource_allocator.cpp (hunk 12 above)
Comment (bechols97) [reply test/resource_allocator.cpp:330]: Added some more tests to cover this and your comment above

File: include/camp/resource.hpp [hunk 13]
```diff
@@ -45,11 +44,11 @@ namespace resources
       template <camp::concepts::ConcreteResource T>
       Resource(T &&value)
       {
-        m_value.reset(new ContextModel<type::ref::rem<T>>(forward<T>(value)));
+        m_value.reset(new ContextModel<camp::decay<T>>(forward<T>(value)));
       }
 
       template <typename T>
-      T* try_get()
+      T *try_get()
```
adayton1: Can you undo all these whitespace changes?

bechols97: Yeah, I can undo them. I ran clang-format from the top level of camp, which caused them.

bechols97: The whitespace changes should be reverted now.

adayton1 (approved):

include/camp/resource.hpp
adayton1: Thanks!

rhornung67 (approved):
Thank you @bechols97

MrBurmark (approved):

adayton1:
Who's going to push the button?

PR #208, opened 2026-07-08
Format code with clang format

bechols97:
Runs clang-format on all headers and source files. This includes the following folders:
* include
* src
* test

This was using the `camp/.clang-format`

### DIFF
--- a/include/camp/concepts.hpp
+++ b/include/camp/concepts.hpp
@@ -256,7 +256,7 @@ namespace concepts
         {arg - increment} -> std::same_as<T>;
         // random access index operator
         arg[increment];
-  };
+    };
 
   template <typename T>
   concept HasBeginEnd = std::ranges::range<T>;
--- a/include/camp/concepts.hpp
+++ b/include/camp/concepts.hpp
@@ -136,7 +136,7 @@ namespace concepts
 
 
     template <typename Ret, typename T>
-    Ret returns(T const&) noexcept;
+    Ret returns(T const &) noexcept;
 
   }  // end namespace detail
 
@@ -146,13 +146,13 @@ namespace concepts
   /// metafunction for use within decltype expression to validate return type is
   /// convertible to given type
   template <typename T, typename U>
-  constexpr auto convertible_to(U&& u) noexcept
-      -> decltype(detail::returns<camp::true_type>(static_cast<T>((U&&)u)));
+  constexpr auto convertible_to(U &&u) noexcept
+      -> decltype(detail::returns<camp::true_type>(static_cast<T>((U &&)u)));
 
   /// metafunction for use within decltype expression to validate type of
   /// expression
   template <typename T, typename U>
-  constexpr auto has_type(U&&) noexcept -> metalib::is_same<T, U>;
+  constexpr auto has_type(U &&) noexcept -> metalib::is_same<T, U>;
 
   template <typename BoolLike>
   constexpr auto is(BoolLike) noexcept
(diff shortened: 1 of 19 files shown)


rhornung67 (dismissed):
Mostly, the position of reference and pointer qualifiers. IMO, a good formatting change.

include/camp/concepts.hpp
```diff
@@ -188,83 +189,99 @@ namespace concepts
   struct requires_ : detail::detected<Op, detail::TL<Args...>> {
   };
 
-  DefineConceptFromSTL(Swappable, std::swappable<T>)
```
adayton1, line 191: Is there a way to tell clang-format not to mess with some code? This section looks significantly worse after formatting.

include/camp/config.in.hpp
```diff
@@ -20,9 +20,9 @@
 #cmakedefine CAMP_WIN_STATIC_BUILD
 #endif
 
-#define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR@
-#define CAMP_VERSION_MINOR @camp_VERSION_MINOR@
-#define CAMP_VERSION_PATCH @camp_VERSION_PATCH@
+#define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR @
```
adayton1: This change needs to be reverted since it is special syntax that CMake is supposed to recognize.

test/array.cpp
```diff
@@ -8,9 +8,9 @@
 //~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~//
 
 #include "camp/array.hpp"
-#include "camp/tuple.hpp"
 
 #include "Test.hpp"
+#include "camp/tuple.hpp"
```
adayton1, line 13: It's strange that this one was moved but not the other camp header.

adayton1 (changes requested):
The only actual bug is the @ symbols being moved in the config.hpp.in header read by CMake.

include/camp/config.in.hpp
rhornung67: Good catch.

include/camp/config.in.hpp
```diff
@@ -20,9 +20,9 @@
...
+#define CAMP_VERSION_MINOR @camp_VERSION_MINOR @
+#define CAMP_VERSION_PATCH @camp_VERSION_PATCH @
```
adayton1:
```suggestion
#define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR@
#define CAMP_VERSION_MINOR @camp_VERSION_MINOR@
#define CAMP_VERSION_PATCH @camp_VERSION_PATCH@
```

include/camp/config.in.hpp
adayton1: Fixed.

adayton1 (dismissed):

include/camp/concepts.hpp
```diff
@@ -188,83 +189,99 @@ namespace concepts
...
+                        { rhs <= lhs } -> std::convertible_to<bool>;
+                        { lhs > rhs } -> std::convertible_to<bool>;
+                        { rhs > lhs } -> std::convertible_to<bool>;
+                        { lhs >= rhs } -> std::convertible_to<bool>;
+                        { rhs >= lhs } -> std::convertible_to<bool>;
+                        { lhs == rhs } -> std::convertible_to<bool>;
+                        { rhs == lhs } -> std::convertible_to<bool>;
+                        { lhs != rhs } -> std::convertible_to<bool>;
+                        { rhs != lhs } -> std::convertible_to<bool>;
+                      };
```
adayton1:
```suggestion
  // clang-format off
  DefineConceptFromSTL(Swappable, std::swappable<T>)
  DefineConceptRequires(LessThanComparable, {arg < arg} -> std::convertible_to<bool>; )
  DefineConceptRequires(GreaterThanComparable, {arg > arg} -> std::convertible_to<bool>; )
  DefineConceptRequires(LessEqualComparable,  {arg <= arg} -> std::convertible_to<bool>; )
  DefineConceptRequires(GreaterEqualComparable,  {arg >= arg} -> std::convertible_to<bool>; )
  DefineConceptRequires(EqualityComparable,  {arg == arg} -> std::convertible_to<bool>; )
  // clang-format on

  template <typename T, typename U>
  concept ComparableTo = requires (T lhs, U rhs) {
        {lhs < rhs} -> std::convertible_to<bool>;
        {rhs < lhs} -> std::convertible_to<bool>;
        {lhs <= rhs} -> std::convertible_to<bool>;
        {rhs <= lhs} -> std::convertible_to<bool>;
        {lhs > rhs} -> std::convertible_to<bool>;
        {rhs > lhs} -> std::convertible_to<bool>;
        {lhs >= rhs} -> std::convertible_to<bool>;
        {rhs >= lhs} -> std::convertible_to<bool>;
        {lhs == rhs} -> std::convertible_to<bool>;
        {rhs == lhs} -> std::convertible_to<bool>;
        {lhs != rhs} -> std::convertible_to<bool>;
        {rhs != lhs} -> std::convertible_to<bool>;
  };
```

include/camp/concepts.hpp
adayton1: See other comment.

include/camp/concepts.hpp
adayton1: I think all the DefineConcept* macro invocations will need the clang-format off/on comments around them.

adayton1: Fixed.

trws: They shouldn't if the semicolon is moved out of the argument.  That's almost certainly what's causing this.  If not, then adding them to one of the lists of macros to handle in the clang-format config should take care of it.

test/array.cpp
trws: It's because the other one has the same name without extension I bet. Odd, but not sure it's worth messing with it.

include/camp/concepts.hpp
adayton1: I'll take a look at that.

adayton1: I think I will just turn off clang-format for now just for these sections. I honestly think the better long term solution is to not use macros for defining these concepts. They hardly save any typing and just obscure the code.

adayton1 (approved):

include/camp/concepts.hpp
trws: Fair. You could always run the header through the preprocessor and take the result if that's your preference.

rhornung67 (approved):
@adayton1 thanks for working through the issues

PR #209, opened 2026-07-08
Remove macro layer in concepts.hpp

adayton1:
The macros added more obscurity than benefit and also were messing up clang-format. Also adds documentation for camp concepts and type traits.

### DIFF
--- a/include/camp/concepts.hpp
+++ b/include/camp/concepts.hpp
@@ -312,6 +312,7 @@ namespace type_traits
   template <class T>
   struct is_iterator : std::bool_constant<camp::concepts::Iterator<T>> {
   };
+
   /// Convenience variable template for `is_iterator<T>::value`.
   template <class T>
   inline constexpr bool is_iterator_v = is_iterator<T>::value;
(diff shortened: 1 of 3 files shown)


include/camp/concepts.hpp
```diff
@@ -70,29 +70,6 @@ namespace concepts
...
-
-#define DefineConceptBinaryAnd(CONCEPT_NAME, expr1, expr2) \
-  template <typename T>                                    \
-  concept CONCEPT_NAME = expr1 && expr2;
-
-#define DefineConceptVar(CONCEPT_NAME, ...) \
-  template <typename T>                     \
-  concept CONCEPT_NAME = EXPAND_AND(__VA_ARGS__);
-
-#define DefineTypeTraitFromConcept(TTName, ConceptName) \
```
bechols97, line 89: Is there any concern with these macros being used in other projects? 
I saw one case in RAJA that would need to be updated. (https://github.com/llnl/RAJA/blob/develop/include/RAJA/pattern/concepts.hpp#L113)

adayton1: These are new since the last release.

bechols97 (dismissed):

johnbowen42 (dismissed):
Great, I did not like these and I'm glad we are simplifying

bechols97 (approved):

include/camp/concepts.hpp
MrBurmark: We should probably take a pass and prefix all the macros with CAMP_ at some point.

trws (approved):
I like it! Now that we don't need the fallback to pre-20 "concepts" this is much nicer without the defines.

PR #212, opened 2026-07-10
Fix "get" function call in tuple header to prevent include ordering issues

gberg617:
The current PR makes get<>(...) unqualified in tuple.hpp invoke_with_order to allow ADL to find the right overload for camp::array, regardless of include ordering.  This PR also adds the array header into camp.hpp.

### DIFF
--- a/include/camp/camp.hpp
+++ b/include/camp/camp.hpp
@@ -21,6 +21,7 @@
 #include "camp/map.hpp"
 #include "camp/number.hpp"
 #include "camp/size.hpp"
+#include "camp/array.hpp"
 #include "camp/tuple.hpp"
 #include "camp/value.hpp"
 
--- a/include/camp/tuple.hpp
+++ b/include/camp/tuple.hpp
@@ -701,7 +701,8 @@ CAMP_HOST_DEVICE constexpr auto invoke_with_order(TupleLike&& tup,
                                                   Fn&& f,
                                                   camp::idx_seq<Sequence...>)
 {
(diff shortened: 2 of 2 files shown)


MrBurmark:
Can you provide some description?

adayton1 (dismissed):

adayton1:
Is there an include order issue?

rhornung67:
Arrghh. So I guess we will need to do a patch release, huh?

gberg617:
> Can you provide some description?

OK, added.

gberg617:
> Is there an include order issue?

Yes.  Please see above description for details.

gberg617:
> Arrghh. So I guess we will need to do a patch release, huh?

camp::array is only used in my SubView PR (https://github.com/llnl/RAJA/pull/1915), so this current fix is only needed for that.

rhornung67:
> > Arrghh. So I guess we will need to do a patch release, huh?
> 
> camp::array is only used in my SubView PR ([llnl/RAJA#1915](https://github.com/llnl/RAJA/pull/1915)), so this current fix is only needed for that.

Thanks for the note @gberg617 We'll probably still do a patch release.

MrBurmark (approved):

MrBurmark:
Hopefully that was the only place we didn't use ADL for these kind of functions.

adayton1 (approved):

gberg617:
> Hopefully that was the only place we didn't use ADL for these kind of functions.

That was the only clear place I found that can benefit from this sort of change.  The other functions in tuple.hpp are specific to tuples.

PR #213, opened 2026-07-15
Fix the GPU Memory Tests

MrBurmark:
Fixed the memory tests that were failing for the gpu backends.

A notable breaking change introduced here is no longer treating the unknown memory type as device in allocate for the GPU resources.

Removed tests that attempted to allocate non-zero amounts of unknown memory type memory.
Make the behavior when explicitly attempting to allocate unknown memory type memory more consistent, all gpu backends now throw if you attempt to allocate unknown memory type memory. Note that the cpu backends completely ignore memory type.

Fix an issue where the Hip resource failed to recognize managed memory pointers. This caused failures during deallocation when the memory type had to be looked up.

Fix an issue where a failed allocation could result in calling memset on nullptr in the calloc implementation.

Use std::calloc in Host::calloc implementation.

Improve unity in code style of allocation and memory functions across resources.


adayton1 (dismissed):

include/camp/resource/sycl.hpp
```diff
@@ -335,52 +335,64 @@ namespace resources
...
-            case MemoryAccess::Pinned:
-              ret = sycl::malloc_host<T>(size, qu);
-              break;
-            case MemoryAccess::Managed:
-              ret = sycl::malloc_shared<T>(size, qu);
-              break;
-          }
+        switch (ma) {
+          case MemoryAccess::Device:
+            ret = sycl::malloc_device<T>(n, qu);
```
artv3, line 346: I think these function can return a nullptr if the system cannot satisfy the request. Do we want to leave it up to the user to check on that?

artv3: maybe a SYCL variant of CAMP_HIP_API_INVOKE_AND_CHECK ?

MrBurmark: I left this returning nullptr as we don't check the return value of malloc which can return nullptr as well.

bechols97 (dismissed):

include/camp/resource/sycl.hpp
artv3: That makes sense. If we want consistent behavior with malloc, should we document that these allocators can return nullptr and that callers are responsible for checking? Or would it be worth adding a helper here for users who want an explicit error path?

MrBurmark: Given that some allocation failures cause exceptions to be thrown and others don't it would probably be better to consistently throw. Let's make an issue to think about this, #217 .

adayton1 (approved):

bechols97 (approved):

include/camp/resource/sycl.hpp
artv3: Thanks!

artv3 (approved):

rhornung67 (approved):
We're still working on polishing the transition to GH Actions and turning off azure checks on another PR branch.

adayton1 (approved):

PR #215, opened 2026-07-17
Transition to Github Actions

rchen20:
Some design choices:
- Azure pipelines are removed, along with old Dockerfile support.
- `rocm6_4_3` and `sycl` GHA jobs now use Radiuss Docker images, like RAJA. `nvcc`'s, and `rocm_latest` still use images distributed from Nvidia and AMD because LLNL does not have those particular images.
- CI test names re-simplified to `build_docker (target)`.


rchen20:
@rhornung67 Just need to change the required checks, and this should be good to go. Thanks!

rhornung67 (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,78 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - uses: threeal/cmake-action@v1.3.0
+        with:
+          cmake-version: '3.29.x'  # Back off version of cmake with known issue
+          build-dir: build
+          options:
+            CMAKE_CXX_STANDARD=20
```
adayton1, line 36: Why not BLT_CXX_STD which sets CMAKE_CXX_STANDARD, CMAKE_CUDA_STANDARD and CMAKE_HIP_STANDARD (the latter two only when building with CUDA or HIP).

PR #218, opened 2026-07-22
Update CAMP_VERSION format

### LINKED ISSUE #216: Version numbers
In camp/config.hpp.in, version numbers are generated like so:

```
#define CAMP_VERSION_MAJOR @camp_VERSION_MAJOR@
#define CAMP_VERSION_MINOR @camp_VERSION_MINOR@
#define CAMP_VERSION_PATCH @camp_VERSION_PATCH@

#define CAMP_VERSION                                           \
  (CAMP_VERSION_MAJOR * 1000000) + (CAMP_VERSION_MINOR * 1000) \
      + (CAMP_VERSION_PATCH)
```

In a concrete example:

```
#define CAMP_VERSION_MAJOR 2025
#define CAMP_VERSION_MINOR 12
#define CAMP_VERSION_PATCH 0
```

This results in CAMP_VERSION being 2025012000, which looks really weird and isn't easy to process mentally. If we took out one order of magnitude for CAMP_VERSION_MAJOR, it would be better (e.g. 202512000). I'd like it even better if it looked like 20251200 or 20260700. To get that, we'd have to change the definition of CAMP_VERSION to the following:

```
#define CAMP_VERSION                                           \
  (CAMP_VERSION_MAJOR * 10000) + (CAMP_VERSION_MINOR * 100) \
      + (CAMP_VERSION_PATCH)
```

The biggest downside would be that you couldn't use code like the following because the version number would actually decrease from 2026.07.0 to 2026.07.1 or anything newer:

```
#if CAMP_VERSION >= 20260701
      m_async_resource.wait_for(m_wait_for_event);
#else
      m_async_resource.wait_for(&m_wait_for_event);
#endif
```

adayton1:
CAMP_VERSION as defined in config.hpp.in would produce version numbers like 2026007000. Since the major and minor release numbers are now based on the year and month and patch releases are generally limited to one or two, a better format would be YYYYMMPP (e.g. 20260700). However, this would be an unexpected change for anyone relying on strictly increasing version numbers for Camp.

Resolves #216.

This is not a blocker for me, so no need to rush this in.


rhornung67 (dismissed):

MrBurmark:
We are getting close to INT_MAX, how about making these `ULL` literals?

adayton1:
> We are getting close to INT_MAX, how about making these `ULL` literals?

The new format is much farther from it, do you still want ULL?

adayton1:
> > We are getting close to INT_MAX, how about making these `ULL` literals?
> 
> The new format is much farther from it, do you still want ULL?

I gave it a try - is that what you were thinking of?

adayton1:
My personal opinion is that this should not be part of a patch release, so I would prefer to wait to merge this. Changing the format from 2026007000 to 20260701 would be very jarring and make using CAMP_VERSION even more complicated with having to handle special cases.

trws:
Agreed. Forcing the change to a ULL seems reasonable, but I would note that even if it didn't have the qualifier on it increasing the version to a number requiring that size would have produced warnings for anyone assigning the literal to a smaller value.  Honestly I'd prefer we keep the format and force the size change rather than change the format, it's less likely to cause downstream breakage that's undetectable by the compiler.

PR #222, opened 2026-08-15
Add Constant Type and Operators

MrBurmark:
Add a constant type that takes a compile-time value as a template parameter. Support arithmetic between constants that produces a constant.

Deprecate integral_constant in favor of constant.

```cpp
constant<0> zero;
constant<1> one;
auto two = one + one;
static_assert(two == 2);
static_assert(std::same_as<decltype(two), constant<2>>);
```

The ultimate goal with this is use it as part of a `RAJA::transparent_range` class that I'm thinking about adding that stores the input begin, end, and stride exactly as they were given. My hope is to improve some cases in RAJAPerf and beyond where the BASE and RAJA differ by knowledge of such exact values. For example we often assume that ranges start at 0 but RAJA ranges always have a runtime begin value, knowledge of which does not cross the host device barrier. This is responsible for some differences between kernel which takes its ranges on the host and launch which takes ranges on the device.

### DIFF
--- a/include/camp/list/list.hpp
+++ b/include/camp/list/list.hpp
@@ -34,7 +34,7 @@ namespace detail
 
   template <typename T, T... Args>
   struct _as_list<int_seq<T, Args...>> {
-    using type = list<integral_constant<T, Args>...>;
+    using type = list<constant<Args>...>;
   };
 }  // namespace detail
 
--- a/include/camp/number.hpp
+++ b/include/camp/number.hpp
@@ -32,43 +32,46 @@ using idx_seq = int_seq<idx_t, vs...>;
 
 namespace detail
 {
-  template <typename T, typename N>
-  struct gen_seq;
-#if CAMP_USE_MAKE_INTEGER_SEQ
   template <typename T, T N>
-  struct gen_seq<T, integral_constant<T, N>> {
+  struct gen_seq;
+
+#if CAMP_USE_MAKE_INTEGER_SEQ
+
+  template <typename T, T N>
+  struct gen_seq {
     using type = __make_integer_seq<int_seq, T, N>;
   };
+
 #elif CAMP_USE_INTEGER_PACK
+
   template <typename T, T N>
-  struct gen_seq<T, integral_constant<T, N>> {
+  struct gen_seq {
     using type = int_seq<T, __integer_pack(N)...>;
   };
+
 #else
+
   template <typename T, typename S1, typename S2>
-  struct concat;
-
+  struct gen_seq_concat;
   template <typename T, T... I1, T... I2>
-  struct concat<T, int_seq<T, I1...>, int_seq<T, I2...>> {
-    using type = typename int_seq<T, I1..., (sizeof...(I1) + I2)...>::type;
+  struct gen_seq_concat<T, int_seq<T, I1...>, int_seq<T, I2...>> {
+    using type = int_seq<T, I1..., (sizeof...(I1) + I2)...>;
   };
 
-  template <typename T, typename N_t>
+  template <typename T>
+  struct gen_seq<T, 0> : int_seq<T> {
+  };
+
+  template <typename T>
+  struct gen_seq<T, 1> : int_seq<T, 0> {
+  };
+
+  template <typename T, T N_t>
   struct gen_seq
-      : concat<T,
-               typename gen_seq<T, integral_constant<T, N_t::value / 2>>::type,
-               typename gen_seq<
-                   T,
-                   integral_constant<T, N_t::value - N_t::value / 2>>::type>::
-            type {
-  };
-
-  template <typename T>
-  struct gen_seq<T, integral_constant<T, 0>> : int_seq<T> {
-  };
-
-  template <typename T>
-  struct gen_seq<T, integral_constant<T, 1>> : int_seq<T, 0> {
+      : gen_seq_concat<T,
+            typename gen_seq<T, N_t / 2>::type,
+            typename gen_seq<T, N_t - N_t / 2>::type
+            >::type {
   };
 #endif
 }  // namespace detail
@@ -77,7 +80,7 @@ namespace detail
 template <idx_t Upper>
 struct make_idx_seq {
   using type =
-      typename detail::gen_seq<idx_t, integral_constant<idx_t, Upper>>::type;
+      typename detail::gen_seq<idx_t, Upper>::type;
 };
 
 // TODO: document
@@ -108,7 +111,7 @@ using idx_seq_from_t = typename idx_seq_from<camp::decay<T>>::type;
 
 // TODO: document
 template <typename T, T Upper>
-struct make_int_seq : detail::gen_seq<T, integral_constant<T, Upper>>::type {
+struct make_int_seq : detail::gen_seq<T, Upper>::type {
 };
 
 // TODO: document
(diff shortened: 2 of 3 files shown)


include/camp/number/number.hpp
```diff
@@ -15,25 +15,276 @@
...
+
+/**
+ * @brief constant yielding logical or for constant values
+ */
+template < auto lhs_value, auto rhs_value >
+CAMP_HOST_DEVICE
+constexpr auto operator||(constant<lhs_value>, constant<rhs_value>)
+{
+  return constant<lhs_value||rhs_value>{};
+}
```
MrBurmark, line 218: This overloads logical and and or so operations between constants remain constants. However this does break short-circuiting if both operands are constants. I'm not sure if this is a big deal or not since presumably constants are generated in constant expressions. Though of course could always return constant<true> from a function with side-effects. So I'm not sure if I'm going to keep these overloads.

MrBurmark: @trws Are there additional concerns related to template instantiations in constant expressions and things like that that short-circuiting also has a bearing on?

rhornung67 (approved):
@MrBurmark I'm OK with this going in for the patch release, but I will leave the final call up to you. If this will require changes to user code, is it to much for a patch release? Also, where should we document those changes?

MrBurmark:
I'll remove the places that this was used in favor or integral_constant for compatibility.

include/camp/number/number.hpp
```diff
@@ -15,25 +15,276 @@
...
+
+/**
+ * @brief constant yielding logical and for constant values
+ */
+template < auto lhs_value, auto rhs_value >
+CAMP_HOST_DEVICE
+constexpr auto operator&&(constant<lhs_value>, constant<rhs_value>)
+{
+  return constant<lhs_value&&rhs_value>{};
+}
```
MrBurmark:
```suggestion
// template < auto lhs_value, auto rhs_value >
// CAMP_HOST_DEVICE
// constexpr auto operator&&(constant<lhs_value>, constant<rhs_value>)
// {
//   return constant<lhs_value&&rhs_value>{};
// }
```

include/camp/number/number.hpp
MrBurmark:
```suggestion
// template < auto lhs_value, auto rhs_value >
// CAMP_HOST_DEVICE
// constexpr auto operator||(constant<lhs_value>, constant<rhs_value>)
// {
//   return constant<lhs_value||rhs_value>{};
// }
```

include/camp/list/list.hpp
```diff
@@ -34,7 +34,7 @@ namespace detail
 
   template <typename T, T... Args>
   struct _as_list<int_seq<T, Args...>> {
-    using type = list<integral_constant<T, Args>...>;
+    using type = list<constant<Args>...>;
```
MrBurmark:
```suggestion
    using type = list<integral_constant<T, Args>...>;
```

include/camp/number/number.hpp
```diff
@@ -15,25 +15,276 @@
...
+
+/**
+ * @brief class that represents a compile time constant value
+ *
+ * TODO: consider making use/match std::integral_constant
+ */
 template <class NumT, NumT v>
-struct integral_constant {
+struct [[deprecated("use constant<val> instead of integral_constant<T, val>")]]
+       integral_constant {
```
MrBurmark:
```suggestion
struct integral_constant {
```

include/camp/number/number.hpp
```diff
@@ -15,25 +15,276 @@
...
+
+
 /**
  * @brief Short-form for a whole number
  *
  * @tparam N The integral value
  */
 template <idx_t N>
-using num = integral_constant<idx_t, N>;
+using num = constant<N>;
```
MrBurmark:
```suggestion
using num = integral_constant<idx_t, N>;
```

include/camp/number/number.hpp
MrBurmark: I went ahead and commented these out.

MrBurmark:
I reverted the places where constant was used inplace of constant so user code doesn't have to change.

include/camp/number/number.hpp
```diff
@@ -15,18 +15,268 @@
...
+
+/**
+ * @brief constant yielding logical and for constant values
+ */
+// template < auto lhs_value, auto rhs_value >
+// CAMP_HOST_DEVICE
+// constexpr auto operator&&(constant<lhs_value>, constant<rhs_value>)
+// {
+//   return constant<lhs_value&&rhs_value>{};
+// }
```
bechols97: Would it be helpful to leave a comment here explaining why this is commented out? (same with `operator||`)

adayton1 (dismissed):

include/camp/number.hpp
```diff
@@ -32,43 +32,46 @@ using idx_seq = int_seq<idx_t, vs...>;
...
   };
 
   template <typename T>
-  struct gen_seq<T, integral_constant<T, 0>> : int_seq<T> {
+  struct gen_seq<T, 1> : int_seq<T, 0> {
   };
 
-  template <typename T>
-  struct gen_seq<T, integral_constant<T, 1>> : int_seq<T, 0> {
+  template <typename T, T N_t>
```
adayton1: I like N better than N_t.

include/camp/number/number.hpp
```diff
@@ -15,18 +15,268 @@
...
+
+/**
+ * @brief constant yielding logical or for constant values
+ */
+// template < auto lhs_value, auto rhs_value >
+// CAMP_HOST_DEVICE
+// constexpr auto operator||(constant<lhs_value>, constant<rhs_value>)
+// {
+//   return constant<lhs_value||rhs_value>{};
+// }
```
MrBurmark:
```suggestion
// Do not overload operator&& and operator|| to allow short-circuiting
// in && and || expressions.
```

adayton1:
Can you update the pull request description now that integral_constant is no longer being deprecated?

include/camp/number/number.hpp
MrBurmark: Just left a comment and removes the operators.

MrBurmark:
> Can you update the pull request description now that integral_constant is no longer being deprecated?

done.

include/camp/number.hpp
MrBurmark: That is reasonable, we might as well do what while we're here.

include/camp/number.hpp
```diff
@@ -32,43 +32,46 @@ using idx_seq = int_seq<idx_t, vs...>;
...
+  struct gen_seq
+      : gen_seq_concat<T,
+            typename gen_seq<T, N_t / 2>::type,
+            typename gen_seq<T, N_t - N_t / 2>::type
```
MrBurmark:
```suggestion
  template <typename T, T N>
  struct gen_seq
      : gen_seq_concat<T,
            typename gen_seq<T, N/2>::type,
            typename gen_seq<T, N - N/2>::type
```

include/camp/number.hpp
MrBurmark: done.

adayton1 (approved):

bechols97 (approved):

rhornung67 (approved):

Issue #223, opened 2026-08-19
Clean up resources


bechols97:
This is bug fix request to add a way to cleanup resources (particularly on Hip/Cuda resources where hipStream/cudaStream types are static created). A couple of options include:

- ~~Making resources owners of streams (this may be be breaking change)~~
- Adding an API call to clean up static resources similar to https://github.com/llnl/RAJA/pull/2078 in RAJA

PR #224, opened 2026-08-21
Adds cleanup functions for resources

bechols97:
This is a bugfix to add cleanup functions for resources.

The Hip and Cuda resources are non-owning and never destroy streams, which show up as leaks in memory checkers/sanitizers. In these cases, the streams would be destroyed in the new cleanup function. If there is no cleanup needed for a resource (such as Host, Omp_target, Sycl), then the cleanup function will be a no-op.

An open question for this PR is whether the cleanup functions should be thread-safe? Currently, these are not thread safe to be consistent with `RAJA::release_unused_internal_memory()`.

Note: These changes were created by AI. I have reviewed the changes and they seemed reasonable.

Related to: #223


rhornung67 (dismissed):
Seems reasonable to me. However, someone else should review and approve to make sure.

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,22 @@ a unique device stream or queue will get a hash. However, since all Host resourc
 one `stream of execution` on the Host), users should be cautious when using the Host resource as a key. For example, 
 ``std::unordered_map<Host, Value> map`` would only ever have one entry.
 
+Cleaning Up Managed Streams
+^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Every concrete resource backend provides a ``cleanup()`` function, and
```
MrBurmark: I would talk about what it does before talking about who has it and if its thread safe.

include/camp/resource/cuda.hpp
```diff
@@ -226,16 +237,50 @@ namespace resources
...
+        return Cuda(nullptr);
 #else
-          CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &s);
+        auto& state = get_stream_state();
+        std::lock_guard<std::mutex> lock(state.mutex);
+        if (state.default_stream == nullptr) {
+          CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate,
+                                         &state.default_stream);
+        }
+        return Cuda(state.default_stream);
```
MrBurmark: Should we try to avoid getting the lock every time we want to create a default stream? Can we use something like a call_once?

docs/sphinx/user_guide/feature/resource.rst
bechols97: I took a second pass at the document. Let me know what you think.

include/camp/resource/cuda.hpp
bechols97: `std::once_flag` that is used by `std::call_once` can not be reset, which means the code can't be reentrant after cleanup. If want this to be reentrant and avoid the lock, then this should probably use `std::atomic_flag` or `std::atomic<bool>` as that can be reset.

MrBurmark: That's true, we can explicitly use the cheap check, lock on failure pattern that I imagine call_once is using.

MrBurmark: I played around with that and the `flag_lock_likely` function in this compiler explorer seems to emit the best code for clang and gcc. https://godbolt.org/z/zGhzdqzxW.

bechols97: Based on our conversation, the `resettable_once_flag` along with using the `flag_lock_likely` implementation for a resettable `call_once` has been added instead of always locking a mutex when getting a stream.

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,39 @@ namespace resources
...
+      struct stream_state {
+        std::array<cudaStream_t, num_streams> streams{};
+        cudaStream_t default_stream = nullptr;
+        int previous = num_streams - 1;
+        camp::resettable_once_flag flag;
+      };
+
+      static stream_state& get_stream_state()
+      {
+        static stream_state state;
```
MrBurmark, line 224: Given that we're manually handling the thread safety now, we should probably make this a class static member variable so it doesn't get another layer of thread safe initialization that comes from being a function local static. Given that we have a library we should probably put the variable definition in the camp sources, as opposed to an inline static member var.

include/camp/resource/cuda.hpp
```diff
@@ -226,16 +239,54 @@ namespace resources
...
+       * Existing resources that refer to CAMP-managed streams are invalid
+       * after this call. Streams passed to CudaFromStream and the CUDA
+       * platform default stream are not destroyed. This function may be
+       * called repeatedly, and later resource construction recreates the
+       * managed streams.
+       *
+       * The caller must ensure no other thread is using CUDA resources while
+       * cleanup runs.
+       */
+      static void cleanup()
```
MrBurmark, line 301: Does it make more sense to put most of this code in a member function of the state?

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,39 @@ namespace resources
...
+            if (s == nullptr) {
+              CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &s);
+            }
+          }
+        });
 
         if (num < 0) {
-          std::lock_guard<std::mutex> lock(s_mtx);
-          s_previous = (s_previous + 1) % num_streams;
-          return s_streams[s_previous];
```
MrBurmark, line 172: This still needs to be locked.

MrBurmark: In fact does it make sense to make this whole function a member function of the state now?

bechols97: It probably makes sense to make getting a stream and getting the default stream member functions at this point.

include/camp/resource/cuda.hpp
bechols97: I can move most of the cleanup logic to be a state member function. However, I think we will still want a static cleanup function on the resource to be consistent across resources.

include/camp/resource/cuda.hpp
bechols97: Moved to logic to member functions and added back guard.

include/camp/resource/cuda.hpp
bechols97: Moved logic to member function

include/camp/resource/cuda.hpp
bechols97: The tradeoff with moving this to a static member variable is that the variable gets constructed pre-main. While the stream_state is trivial constructible/destructible, I am not sure we want this to be initialized pre-main. But, if that isn't a concern, I can make this change.

MrBurmark: The reason I was hoping to make this in a global instead of a local static is that local statics get the initialization logic and we are now handling explicitly, so it would be doubled. Perhaps we could use `constinit` to have the compiler check that we aren't doing any dynamic initialization and convince it to not emit those checks?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include <utility>
+
+namespace camp
+{
+
+/// Resettable version of std::once_flag
+///
+/// This is similar to std::once_flag used in std::call_once. However,
+/// this version supports resetting flag, so that memory can be cleaned
+/// and re-initialized at a later if needed.
```
MrBurmark:
```suggestion
/// and re-initialized at later if needed.
```

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,92 @@
...
+/// 
+/// @note Currently, this uses atomic<bool> instead of atomic_flag to support
```
MrBurmark, line 26: Checking `is_always_lock_free` is not a bad idea but we could also emit a warning by using the macro version that I assume is `ATOMIC_BOOL_LOCK_FREE` as described here https://en.cppreference.com/cpp/atomic/atomic/is_always_lock_free.

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,92 @@
...
+/// data to be cleaned up. If needed, calling this function again after
+/// resetting the flag will call the callable function again.
+template <typename Callable, typename... Args>
+void call_once(camp::resettable_once_flag& flag, Callable&& callable, Args&&... args)
+{
+  if (flag.test(std::memory_order::acquire)) [[likely]] {
+    return;
+  }
+
+  std::lock_guard guard(flag.m_lock);
```
MrBurmark: Did this need to be a friend if `get_mutex` exists?

test/resource.cpp
```diff
@@ -1881,3 +1881,118 @@ TEST(CampResourceSycl, Helpers)
...
+{
+  Host::cleanup();
+#ifdef CAMP_HAVE_CUDA
+  {
+    Cuda cuda_resource;
+    Cuda cuda_default = Cuda::get_default();
+    cuda_resource.wait();
+    cuda_default.wait();
+  }
+#endif
```
MrBurmark: How about doing this in a template function/lambda so the same things is done for all resources?

test/resource.cpp
```diff
@@ -1881,3 +1881,118 @@ TEST(CampResourceSycl, Helpers)
...
+    Hip recreated_hip_default = Hip::get_default();
+    recreated_hip_resource.wait();
+    recreated_hip_default.wait();
+  }
+  Hip::cleanup();
+#endif
+}
+
+#ifdef CAMP_HAVE_CUDA
+TEST(CampResourceCuda, CleanupPreservesCustomStream)
```
MrBurmark: Is there an equivalent sycl or openmp test for this?

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,82 @@ namespace resources
...
-            CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &s);
+      static constexpr int num_streams = 16;
+
+      class stream_state {
+      public:
+        cudaStream_t get_default_stream()
+        {
+#if !CAMP_USE_PLATFORM_DEFAULT_STREAM
+          auto& default_stream = m_default_stream;
+          camp::call_once(m_default_flag, [&default_stream] () {
```
MrBurmark: Would explicit `this` capture be preferable to making a local reference?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,92 @@
...
+  if (!flag.test(std::memory_order::relaxed)) {
+    callable(std::forward<Args>(args)...);
+    flag.set(true, std::memory_order::release);
```
MrBurmark, line 82: Do we have to worry about potential exceptions when calling the callable? I guess if it throws then the flag won't get set and that's the behavior we want anyway.

MrBurmark (changes requested):

include/camp/init_helpers.hpp
bechols97: I don't think we have to worry about exceptions when calling the callable. As you mentioned, the flag won't get set if an exception is thrown, and the lock guard will unlock the mutex when out of scope so we don't have to worry about a deadlock.

include/camp/resource/cuda.hpp
bechols97: Moved to explicit `this` capture

test/resource.cpp
bechols97: Setup template function to reuse code for all resources

include/camp/init_helpers.hpp
bechols97: Good catch, originally the `get_mutex` didn't exist and wasn't needed outside of `camp::call_once`. The friend has been removed

include/camp/init_helpers.hpp
MrBurmark: Ok, if it calls it multiple times and fails multiple times, what can you do.

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,92 @@
...
+#ifndef CAMP_INIT_HELPERS_HPP
+#define CAMP_INIT_HELPERS_HPP
+
+#include <atomic>
+#include <mutex>
+#include <utility>
+
+namespace camp
+{
+#if !ATOMIC_BOOL_LOCK_FREE
```
MrBurmark: You double checked that this is defined on all of our platforms?

include/camp/resource/cuda.hpp
MrBurmark: we should probably try this out on compiler explorer to see what all happens either way.

include/camp/init_helpers.hpp
bechols97: Unfortunately, the `#warning <message>` isn't defined with MSVC. Apparently, this a compiler extension in gcc and clang until c++23. For now, I just removed this extra check.

MrBurmark: oh well, we can survive without it.

include/camp/resource/cuda.hpp
bechols97: Moved to use `constinit` to avoid any thread synchronization with the local static variable as we discussed.

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,85 @@ namespace resources
...
+          camp::resettable_once_flag flag;
+          std::array<cudaStream_t, num_streams> streams{nullptr};
+          int previous{num_streams - 1};
+        };
+
+        alignas(camp::hardware_destructive_interference_size) default_state m_default;
+        alignas(camp::hardware_destructive_interference_size) extra_state m_extra;
+      };
+
+      static constinit stream_state streams;
```
MrBurmark, line 230: Can we just put inline here and avoid the definition below?

include/camp/resource/sycl.hpp
```diff
@@ -308,6 +308,9 @@ namespace resources
       // get default resource
       static Sycl get_default() { return Sycl(0, get_default_context()); }
 
+      /// Clean up resources managed by the SYCL backend.
+      static void cleanup() {}
```
MrBurmark: Do we not want to setup sycl similarly to cuda/hip so users can destroy the sycl objects early?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,101 @@
...
+///
+/// This is similar to std::once_flag used in std::call_once. However,
+/// this version supports resetting flag, so that memory can be cleaned
+/// and re-initialized later if needed.
+/// 
+/// @note Currently, this uses atomic<bool> instead of atomic_flag to support
+///       older versions of GCC. atomic_flag::test is not supported until
+///       GCC 11. atomic_flag would be preferred as it is guaranteed to
+///       not use a lock.
+class alignas(camp::hardware_constructive_interference_size)
```
MrBurmark: We probably don't want the whole class to be aligned. I think that affects its size.

include/camp/resource/cuda.hpp
bechols97: Unfortunately, no. The compiler ran into error out due to the outer class Hip/Cuda not being fully defined. I believe this is due to constinit initializing the object at compiler time, so the compiler needs the class to be fully defined.

include/camp/init_helpers.hpp
bechols97: Ah ok, that would be problematic if it affects the size. Removed this

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
 one `stream of execution` on the Host), users should be cautious when using the Host resource as a key. For example, 
 ``std::unordered_map<Host, Value> map`` would only ever have one entry.
 
+Cleaning Up Resources
+^^^^^^^^^^^^^^^^^^^^^
+
+The ``cleanup`` function will delete/destroy any Camp-managed global state
+that is used by resources. For example, this includes Cuda/Hip streams managed
```
artv3:
```suggestion
that is used by resources. For example, this includes CUDA/HIP streams managed
```
Capital letters since they are acronyms

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
 one `stream of execution` on the Host), users should be cautious when using the Host resource as a key. For example, 
 ``std::unordered_map<Host, Value> map`` would only ever have one entry.
 
+Cleaning Up Resources
+^^^^^^^^^^^^^^^^^^^^^
+
+The ``cleanup`` function will delete/destroy any Camp-managed global state
```
artv3:
```suggestion
The ``cleanup`` function will delete any Camp-managed global state
```
delete and destroy are kind of the same thing, we can probably just use one

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
...
+that is used by resources. For example, this includes Cuda/Hip streams managed
+by Camp. Every concrete resource backend provides a ``cleanup()`` function,
+and ``camp::resources::cleanup()`` calls ``cleanup()`` for every enabled backend.
+Resources that do not currently require explicit runtime destruction
+will have no-ops for their respective cleanup functions. 
+
+Cleanup invalidates every existing resource that refers to a Camp-managed
+stream. Applications must finish using those resources and ensure that no
+other thread is using them before cleanup. A later resource construction will
```
artv3:
```suggestion
other thread is using them before cleanup. Constructing a resource later. construction will
```

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
...
+recreate the managed streams, but it does not make an older resource valid
+again.
+
+.. note::
+   All ``cleanup()`` functions are not thread-safe.
```
artv3:
```suggestion
   The ``cleanup()`` functions are not thread-safe.
```

artv3 (commented):
minor doc suggestions

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,85 @@ namespace resources
...
     class Cuda
     {
-      static cudaStream_t get_a_stream(int num)
-      {
-        static constexpr int num_streams = 16;
-        static std::array<cudaStream_t, num_streams> s_streams = [] {
-          std::array<cudaStream_t, num_streams> streams;
-          for (auto& s : streams) {
-            CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &s);
+      static constexpr int num_streams = 16;
```
artv3, line 156: This seems like a magic number at first glance, maybe a more descriptive name?

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,85 @@ namespace resources
...
+          camp::call_once(m_default.flag, [this] () {
+            if (m_default.stream == nullptr) {
+              CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &m_default.stream);
+            }
+          });
+#endif
+          return m_default.stream;
+        }
+
+        cudaStream_t get_a_stream(int num)
```
artv3:
```suggestion
        cudaStream_t get_a_stream(int index)
```
for clarity?

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,85 @@ namespace resources
...
+        {
+          camp::call_once(m_extra.flag, [this] () {
+            for (auto& s : m_extra.streams) {
```
artv3:
```suggestion
            for (auto& stream : m_extra.streams) {
```
for clarity?

include/camp/resource/cuda.hpp
bechols97: Would renaming this to `NUM_STREAMS` be more clear? Or, did you have another variable name in mind?

docs/sphinx/user_guide/feature/resource.rst
bechols97: I rephrased this part to clarify. Let me know if there are other suggestions to the documentation.

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,100 @@
...
+/// NVCC doesn't define interference size with device compile
+/// and gets a compiler error. This adds default values to 
+/// avoid this compiler error.
+#ifdef __cpp_lib_hardware_interference_size
+  using std::hardware_constructive_interference_size;
+  using std::hardware_destructive_interference_size;
+#else
+  inline constexpr std::size_t hardware_constructive_interference_size = 64;
+  inline constexpr std::size_t hardware_destructive_interference_size = 64;
+#endif
```
MrBurmark: This seems like it will end up being a potential ABI break for types used on the host and device. Perhaps it would be better to not worry about this for now and allow multiple once flags per cache line like the compilers do for init guards.

bechols97: Good catch, removed this for now.

include/camp/resource/cuda.hpp
MrBurmark: This magic number was here already, we can think about changing it later if necessary.

include/camp/resource/sycl.hpp
```diff
@@ -190,106 +190,107 @@ namespace resources
...
+          }
+          syclContext = &get_thread_default_context();
 
-          if (cachedContextIter == queueMap_end) {
-            cachedContextIter = queueMap.find(syclContext);
-            if (cachedContextIter == queueMap_end) {
-              static constexpr auto gpuSelector = sycl::gpu_selector_v;
-              static const sycl::property_list propertyList =
-                  sycl::property_list(sycl::property::queue::in_order());
+          auto cachedContextIter = get_cache_context();
```
MrBurmark:
```suggestion
          auto& cachedContextIter = get_cache_context();
```

include/camp/resource/sycl.hpp
```diff
@@ -190,106 +190,107 @@ namespace resources
...
 
+          if (syclContext != cachedContextIter->first) {
```
MrBurmark, line 227: Doesn't `cachedContextIter` always start at end which is not de-reference-able?

include/camp/resource/sycl.hpp
```diff
@@ -190,106 +190,107 @@ namespace resources
...
-          }
 
-          if (num < 0) {
-            int& previous = cachedContextIter->second.first;
-            previous = (previous + 1) % num_queues;
-            return cachedContextIter->second.second[previous];
+            if (index < 0) {
+              int& previous = cachedContextIter->second.previous;
+              previous = (previous + 1) % num_queues;
+              return cachedContextIter->second.queues[previous];
```
MrBurmark, line 261: Were we going to lock this?

include/camp/resource/sycl.hpp
bechols97: Yes, that always starts at the end. This was the condition that was there previously, but we probably want an extra check to verify `cachedContextIter != m_queue_map.end()`.

include/camp/resource/sycl.hpp
bechols97: After reviewing the code, the lock guard is at the top of the outer conditional. So, if we lock again here, it deadlocks.

include/camp/resource/sycl.hpp
```diff
@@ -190,106 +190,107 @@ namespace resources
...
+        using queue_map_type      = std::map<const sycl::context*, queue_list>;
+        using queue_map_iter_type = queue_map_type::iterator;
+
+      public:
+        static constexpr int num_queues    = 16;
+        static constexpr auto gpu_selector = sycl::gpu_selector_v;
+
+        auto& get_cache_context()
+        {
+          thread_local auto cachedContextIter = m_queue_map.end();
```
MrBurmark, line 207: Do you need to reset this or any other such state like the context stuff above as part of cleanup?

include/camp/resource/sycl.hpp
MrBurmark: I was worried about end potentially becoming invalid, but I guess this is a map so iterators only get invalidated if their data gets removed.

include/camp/resource/sycl.hpp
MrBurmark: Oh, I didn't notice this was in that scope.

include/camp/resource/sycl.hpp
MrBurmark: yep that should be fixed.

include/camp/resource/sycl.hpp
MrBurmark, line 120: If we want things to be reusable after cleanup we should use the resettable_once_flag for these.

MrBurmark: I wonder if would simplify things even a bit more to make a templated class that combined a resettable_once_flag with the object and some lifetime methods? Basically a std optional with thread safe initialization that is intended as a reset-able replacement for the scott meyers singleton pattern.

MrBurmark: Then we could just have constinit static or constinit thread_local variables and call something like init_once and reset on them.
```cpp
constinit static camp::optional_singleton<MyType> singleton;
singleton.emplace_once(constructor, args);
singleton.reset();
```

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,88 @@
...
+    m_flag.store(false, std::memory_order_release);
+  }
+
+  std::mutex& get_mutex() noexcept
+  {
+    return m_lock;
+  }
+
+private:
+  std::mutex m_lock{};
```
MrBurmark, line 64: Take a look at atomic wait and notify_all, we may not need a separate mutex.

bechols97: Optimizing the `camp::call_once` and `camp::resettable_once_flag` further could be done in a later PR. Additionally, @MrBurmark mentioned that wait and notify_all are added in gcc version 11.

include/camp/resource/sycl.hpp
bechols97: I added a `camp::optional_singleton` type to handle this. There was a weird segfault at the end of main with the previous static context. Moving the static context to the global `Sycl::queue_state` fixed this and all tests were passing. 

I haven't added tests for the `camp::optional_singleton`, but I plan to add tests for the `optional_singleton` if the interface looks reasonable. Also, does `optional_singleton` seem like a good name, or should this be `resettable_singleton` or something else?

include/camp/resource/sycl.hpp
```diff
@@ -106,6 +107,126 @@ namespace resources
...
+          constinit thread_local camp::optional_singleton<sycl::context> t_context;
+          return t_context;
+        }
+
+        auto& get_cache_context()
+        {
+          constinit thread_local camp::optional_singleton<queue_map_iter_type> cachedCtxIterManager;
+          cachedCtxIterManager.emplace_once(m_queue_map.end());
+          return cachedCtxIterManager; 
+        }
```
MrBurmark: These should probably be non-member vars now that they can be.

include/camp/resource/sycl.hpp
```diff
@@ -116,9 +237,15 @@ namespace resources
       static sycl::context& get_private_context(
           const sycl::context* syclContext)
       {
-        static sycl::context s_context(syclContext ? *syclContext
-                                                   : sycl::context());
+#if 0
```
MrBurmark: What's up with the ifdefs?

bechols97: Oh sorry, I left in some testing code. I will remove

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,172 @@
...
+/// \note This allows constant initialization (initialization at compiler time)
+/// for all types due the construction of the \tparam{T} happening at a later time.
+/// 
+template <typename T, OptionalDtorPolicy Policy = OptionalDtorPolicy::Default>
+class optional_singleton
+{
+public:
+  optional_singleton() = default;
+
+  ~optional_singleton() requires (Policy == OptionalDtorPolicy::None) = default;
```
MrBurmark: Perhaps we should add or T is trivially destructible?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  void reset() noexcept
+  {
+    if (has_value()) { std::destroy_at(reinterpret_cast<T*>(m_storage)); }
+    m_flag.clear(); 
+  }
+
+private:
+  alignas(alignof(T)) unsigned char m_storage[sizeof(T)]{};
```
MrBurmark, line 166: Do we want the {}?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  constexpr const T& value() const
+  {
+    return *std::launder(reinterpret_cast<const T*>(m_storage));
+  }
+
+  constexpr bool has_value() const noexcept
+  {
+    return m_flag.test(std::memory_order_relaxed);
```
MrBurmark: Should this be test acquire?

include/camp/resource/sycl.hpp
```diff
@@ -106,6 +107,119 @@ namespace resources
...
+
+      struct context
+      {
+        inline constinit static singleton_t<sycl::context> default_ctx;
+
+        static auto& get_thread_private_context()
+        {
+          constinit thread_local singleton_t<sycl::context> t_context;
+          return t_context;
+        }
```
MrBurmark, line 121: Does this have to be in a function?

MrBurmark: Did we decide that we needed more ordering?

include/camp/resource/sycl.hpp
bechols97: `constinit thread_local` non-member variables fail to compile.

include/camp/resource/sycl.hpp
bechols97: `constinit thread_local` non-member variables fail to compile. So, this has to be in a function. I am still working out some weird destructor ordering issues, so there will be some additional changes.

include/camp/init_helpers.hpp
bechols97: The `{}` is required here because `resettable_once_flag` not being trivially constructible and all members need to be initialized to have a constant initializer. I can technically remove that, but then it requires the {} anytime that `constinit optional_singleton` is used, which seemed less ideal.

include/camp/init_helpers.hpp
bechols97: Updated

include/camp/init_helpers.hpp
bechols97: Added is_trivially_destructible

include/camp/resource/sycl.hpp
MrBurmark: what is it complaining about?

include/camp/init_helpers.hpp
MrBurmark: ah, that's a bit annoying but understandable.

include/camp/resource/cuda.hpp
```diff
@@ -152,27 +153,85 @@ namespace resources
...
-        static constexpr int num_streams = 16;
-        static std::array<cudaStream_t, num_streams> s_streams = [] {
-          std::array<cudaStream_t, num_streams> streams;
-          for (auto& s : streams) {
-            CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamCreate, &s);
+      static constexpr int num_streams = 16;
+
+      class stream_state {
+      public:
+        cudaStream_t get_default_stream()
```
MrBurmark, line 160: did we want to use the singleton class for cuda/hip?

include/camp/resource/sycl.hpp
bechols97: The error message was "non-static data member '<variable name>' declared 'constinit'". I don't think it is possible to have a `thread_local` non-member variable.

bechols97:
The sycl resource was causing some strange behavior after main when trying to use the `optional_singleton` class. So, I reverted the changes in the sycl backend and plan to revisit this later. Regarding the hip/cuda resources, I moved these to use the `optional_singleton`.

include/camp/resource/cuda.hpp
bechols97: Updated to use the singleton class for both cuda/hip

include/camp/resource/sycl.hpp
bechols97: I will take another pass at cleaning up the Sycl resource after this PR gets merged.

MrBurmark: Did you try `inline static thread_local constinit`?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,116 @@ namespace resources
...
-        static std::mutex s_mtx;
-        static int s_previous = num_streams - 1;
+      inline static constinit singleton_t<HipStream> default_stream;
+      inline static constinit singleton_t<ExtraStream> extra_streams;
+
+      static cudaStream_t get_default_stream()
+      {
+#if !CAMP_USE_PLATFORM_DEFAULT_STREAM
+        default_stream.emplace_once();
+        return default_stream.value().get_handle();
```
MrBurmark:
```suggestion
        return default_stream->get_handle();
```
Would this be a clearer usage pattern here?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,116 @@ namespace resources
...
+#else
+        return nullptr;
+#endif
+      }
+
+      static cudaStream_t get_a_stream(int num)
+      {
+        extra_streams.emplace_once();
+        auto& extra_state = extra_streams.value();
```
MrBurmark:
```suggestion
        auto& extra_state = extra_streams.get_or_emplace();
```
Would this be a clearer API?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,116 @@ namespace resources
       }
     };
 
+    class CudaStream
+    {
+    public:
```
MrBurmark:
```suggestion
    struct CudaStream
    {
```
Would this be a bit more clear?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,116 @@ namespace resources
...
+      CudaStream& operator=(CudaStream const&) = delete;
+
+      CudaStream& operator=(CudaStream&& rhs) noexcept
+      {
+        finalize(m_stream);
+        m_stream = std::exchange(rhs.m_stream, nullptr);
+        return *this;
+      }
+
+      ~CudaStream() { finalize(m_stream); }
```
MrBurmark: Do we also want a way to manually finalize?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,116 @@ namespace resources
...
+
+      Platform get_platform() const { return Platform::cuda; }
+
+      void wait() const
+      {
+        CAMP_HIP_API_INVOKE_AND_CHECK(cudaStreamSynchronize, m_stream);
+      }
+
+      handle_type get_handle() const { return m_stream; }
```
MrBurmark: Do we want to have a way to check if this has a value, like `empty` or `has_value` or `explicit operator bool`?

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,172 @@
...
+/// \note This allows constant initialization (initialization at compiler time)
+/// for all types due the construction of the \tparam{T} happening at a later time.
+/// 
+template <typename T, OptionalDtorPolicy Policy = OptionalDtorPolicy::Default>
+class optional_singleton
+{
+public:
+  optional_singleton() = default;
+
+  ~optional_singleton() requires (Policy == OptionalDtorPolicy::None || std::is_trivially_destructible_v<T>) = default;
```
MrBurmark: If we're going to repeat this logic maybe put this into a private static constexpr bool?

include/camp/resource/cuda.hpp
bechols97: Yeah, it probably would be better to support `operator->`. Looking at https://en.cppreference.com/cpp/utility/optional, were there any additional `std::optional` member functions that we want to support in this type?

include/camp/resource/cuda.hpp
bechols97: The `emplace_once` returns the value as well to match `std::optional`. So, `auto& extra_state = extra_streams.emplace_once();` will produce the same behavior as the `get_or_emplace()` function in your example. Do you think having a `get_or_emplace` is more readable than just updating to `auto& extra_state = extra_streams.emplace_once();`?

include/camp/resource/cuda.hpp
bechols97: I don't have a strong preference on this using `struct` vs. `class`. The `CudaStream`/`HipStream` objects are trying to follow `CudaEvent`/`HipEvent` with being a move only `class` and owning a stream. The reason for these wrappers is for the `optional_singleton`, which needs a constructor and destructor.

include/camp/resource/cuda.hpp
MrBurmark: I figured it would be good to hit the core set to make things a bit easier to use and manage lifetime wise if we're going to allow an "empty" state.

include/camp/resource/cuda.hpp
MrBurmark: If we're often using the return value then get_or_emplace seems a bit more direct.

include/camp/resource/cuda.hpp
rhornung67: @bechols97 I think the @MrBurmark 's point is that methods are default `public` for `struct` and default `private` for `class`. That's really the main difference here.

MrBurmark: Ya, this just saves a line, its not a big deal either way.

include/camp/resource/cuda.hpp
MrBurmark: I guess this is less necessary if the only way to get here is via move.

bechols97: If we add a way to manually cleanup the stream, then we will like want some way to check if the stream is valid. If not, we could likely assume that the stream is valid for the lifetime of the object.

MrBurmark: Ya, since we're destroying this via the singleton its perhaps not as critical to have it in the stream wrapper class. If we ever decide to generalize the stream wrapper to be more like unique_ptr and make it possible to give a stream to the class or steal a stream from the class then we'll need more, but its probably not necessary now.

include/camp/resource/cuda.hpp
bechols97: Added `get_or_emplace` along with some tests

include/camp/resource/cuda.hpp
bechols97: Updated with extra member functions to more closely match `std::optional`. I intentionally missed `value_or` since users likely meant `get_or_emplace` in the `optional_singleton` case. However, let me know if I missed something.

include/camp/resource/cuda.hpp
bechols97: Moved to `struct`

include/camp/resource/cuda.hpp
bechols97: I will leave this out for now. If need be, I can add it back in later.

include/camp/init_helpers.hpp
bechols97: Moved the common logic to a private static constexpr bool.

include/camp/resource/cuda.hpp
bechols97: Correct me if I am wrong, but I think in another conversation we decided to not have a manual finalize as it is not currently needed and could update this later if the need comes up.

MrBurmark: ya, let's leave that for later.

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,113 @@ namespace resources
...
+      }
+
+      ~CudaStream() { finalize(m_stream); }
+
+      Platform get_platform() const { return Platform::cuda; }
+
+      void wait() const
+      {
+        CAMP_CUDA_API_INVOKE_AND_CHECK(cudaStreamSynchronize, m_stream);
+      }
```
MrBurmark: Are these used? Should this be paired down to what is actually used?

include/camp/resource/cuda.hpp
```diff
@@ -150,29 +151,113 @@ namespace resources
...
+    };
+
     class Cuda
     {
-      static cudaStream_t get_a_stream(int num)
+      static constexpr int num_streams = 16;
+
+      template <typename T>
+      using singleton_t = camp::optional_singleton<T, camp::OptionalDtorPolicy::None>;
```
MrBurmark: Comment on why we're not destroying these?

include/camp/resource/sycl.hpp
MrBurmark: Should we comment here that there are resources to cleanup but its not implemented yet?

include/camp/resource/sycl.hpp
MrBurmark: Maybe `resettable_singleton` is a bit better.

include/camp/resource/cuda.hpp
bechols97: Added comment

include/camp/resource/cuda.hpp
bechols97: Removed unused features

include/camp/resource/sycl.hpp
bechols97: Renamed to `resettable_singleton`

include/camp/resource/sycl.hpp
bechols97: Added comment to mention this is not implemented yet

MrBurmark (dismissed):

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
...
 ``std::unordered_map<Host, Value> map`` would only ever have one entry.
 
+Cleaning Up Resources
+^^^^^^^^^^^^^^^^^^^^^
+
+The ``cleanup`` function will delete any Camp-managed global state
+that is used by resources. For example, this includes CUDA/HIP streams managed
+by Camp. Every concrete resource backend provides a ``cleanup()`` function,
+and ``camp::resources::cleanup()`` calls ``cleanup()`` for every enabled backend.
+Resources that do not currently require explicit runtime destruction
```
adayton1: It might be worth calling out that sycl cleanup is not supported.

docs/sphinx/user_guide/feature/resource.rst
```diff
@@ -183,3 +183,21 @@ a unique device stream or queue will get a hash. However, since all Host resourc
...
+will have no-ops for their respective cleanup functions. 
+
+Cleanup invalidates every existing resource that refers to a Camp-managed
+stream. Applications must finish using those resources and ensure that no
+other thread is using them before cleanup. If an application uses creates
```
adayton1: I think there's a typo in here, and probably some additional clarification would be helpful.

include/camp/resource/host.hpp
```diff
@@ -93,6 +93,9 @@ namespace resources
         return h;
       }
 
+      /// Clean up resources managed by the host backend.
```
adayton1: Does this mean there are no resources to be cleaned up? If so, document that.

include/camp/resource/omp_target.hpp
```diff
@@ -181,6 +181,9 @@ namespace resources
         return o;
       }
 
+      /// Clean up resources managed by the OpenMP target backend.
```
adayton1: Does this mean there are no resources to be cleaned up? If so, document that.

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,200 @@
...
+/// will call the deconstructor.
+enum class DestructorPolicy
+{
+  Default,
+  None
+};
+
+/// Resettable version of a singleton
+///
+/// This class aims to have clear semnatics around initialization and
```
adayton1: Typo: semnatics

include/camp/init_helpers.hpp
```diff
@@ -0,0 +1,200 @@
...
+/// destruction of singleton types. Types will not be constructed until
+/// `emplace_once` is called. Additionally, if need be types, can be
```
adayton1: I'm not following this last sentence.

include/camp/init_helpers.hpp
bechols97: Fixed typo

include/camp/init_helpers.hpp
bechols97: I tried to clarify this. Let me know if it makes more sense now.

include/camp/resource/host.hpp
bechols97: Documented that there is nothing to be cleaned up here.

include/camp/init_helpers.hpp
adayton1: That's better, thanks!

include/camp/resource/omp_target.hpp
bechols97: Documented that there is nothing to be cleaned up here.

docs/sphinx/user_guide/feature/resource.rst
bechols97: I went rewrote this section to hopefully better clarify. Let me know if the current version is better or if it needs another pass.

docs/sphinx/user_guide/feature/resource.rst
bechols97: Added a note for the sycl backend not being supported.

adayton1 (approved):

MrBurmark (approved):

PR #227, opened 2026-08-31
add specialization for zero-sized array

gberg617:


include/camp/array.hpp
```diff
@@ -192,6 +192,134 @@ struct array {
...
+
+template <class T>
+CAMP_HOST_DEVICE inline constexpr bool operator!=(const array<T, 0>&,
+                                                  const array<T, 0>&)
+{
+  return false;
+}
+
+template <class T>
+CAMP_HOST_DEVICE inline constexpr bool operator<(const array<T, 0>&,
```
gberg617, line 296: Note: C++20 replaces most of these comparison operators with the three-way comparison `<=>` operator

test/array.cpp
adayton1, line 253: Can add a check on the return value of empty() for a zero length array?

adayton1: Or if you already check it above, just remove this comment.

adayton1, line 238: Can you add a check for size on zero length array? Or if that's already checked above, just remove this comment.

gberg617: I already added a test that checks empty() so I just removed the comments and re-named this test "non_empty" to be more accurate

gberg617: Also checked already, so I just removed the comment.

adayton1 (approved):

MrBurmark (approved):

